PR #28, opened 2018-04-14
1.0.0

daedalus28:
Prepare for new things that were developed in contexture-imdb for convenience


src/exampleTypes/ResultPager.js
```diff
@@ -0,0 +1,69 @@
...
+import InjectTreeNode from '../utils/injectTreeNode'
+
+// These are to prevent warning from `active`, `previous`, `next`
+let span = ({ children }) => <span>{children}</span>
+let a = ({ children, onClick }) => <a onClick={onClick}>{children}</a>
+
+export default InjectTreeNode(
+  observer(({ node, tree, List = 'div', Item = span, Link = a }) => {
+    let pages = Math.ceil(
+      (node.context.response.totalRecords || 1) / node.pageSize
```
sadasant: Is it safe to assume that `node.context.response.totalRecords` exists? also, would it make sense to have a default value for pageSize?

src/exampleTypes/ResultPager.js
```diff
@@ -0,0 +1,69 @@
...
+                  }
+                >
+                  {page + (n + 1)}
+                </Link>
+              </Item>
+            ),
+          2
+        )}
+        <Item disabled={!(page < pages)}>
+          <Link next onClick={() => tree.mutate(node.path, { page: page + 1 })}>
```
sadasant: If we make these functions receive the `node` directly, we could declare them outside of the component and prevent a re-generation of the function on re-render.

src/exampleTypes/ResultTable.js
```diff
@@ -0,0 +1,109 @@
...
+            </tr>
+          </thead>
+          <tbody>
+            {_.map(
+              x => (
+                <tr key={x._id}>
+                  {_.map(
+                    ({ field, display, Cell = 'td' }) => (
+                      <Cell key={field}>
+                        {(display || (x => x))(
```
sadasant: Maybe this default value could be a default value of the display property received in this map function.

src/exampleTypes/Styles.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+
+export default () => (
+  <style>
+    {`.contexture-search-box:focus {
+      background: rgba(255, 255, 255, 1);
```
sadasant: This is white, no? 🤔

src/layout/Awaiter.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+import { observer } from 'mobx-react'
+
+export default observer(({ promise, children }) =>
+  promise.case({
+    pending: () => <div>Loading...</div>,
+    rejected: error => <div>Ooops.. {error}</div>,
+    fulfilled: value => <div>{children(value)}</div>,
```
sadasant: I'm afraid that having mixed children as functions and as children of nodes and as children of components can be confusing.

src/layout/Flex.js
```diff
@@ -0,0 +1,6 @@
+import React from 'react'
+
+// TODO: align items/content/justify/etc support
+export let Flex = ({ style, ...props }) => (
+  <div style={{ display: 'flex', ...style }} {...props} />
+)
```
sadasant: We can pretty much make this better with out marketplace's Flex thing.

src/layout/Pickers.js
```diff
@@ -0,0 +1,59 @@
...
+      <div>
+        <Input {...value(filter)} />
+        {_.map(
+          ({ value, label }) => (
+            <Item key={value} onClick={() => onChange(value)}>
+              <Highlight text={label} pattern={F.view(filter)} />
+            </Item>
+          ),
+          _.filter(x => F.matchAllWords(F.view(filter))(x.label), options)
+        )}
```
sadasant: From a performance standpoint, maybe these maps should be separate components that receive just what they need. For example, here, this component will receive:
- `_.filter(x => F.matchAllWords(F.view(filter))(x.label), options)`,
- onChange.

That way, React won't have to re-run the map function if these properties haven't changed.

sadasant: If this idea is interesting to you, there are many other places where we could re-think small things this way.

src/layout/Popover.js
```diff
@@ -6,8 +6,8 @@ let { fullscreen } = styles
 
 // Simple popover
 export default observer(
-  ({ show, children, style }) =>
-    f.view(show) && (
+  ({ isOpen, children, style }) =>
+    f.view(isOpen) && (
```
sadasant: Let's make sure futil is used with the upper case `F` even in this file. Several lines more in the diff won't matter at this point.

src/exampleTypes/Facet.js
```diff
@@ -0,0 +1,53 @@
...
+    <div {...props}>
+      {!hide.facetFilter && (
+        <input
+          className="contexture-search-box"
+          type="text"
+          value={node.optionsFilter}
+          onChange={e =>
+            tree.mutate(node.path, { optionsFilter: e.target.value })
+          }
+          placeholder="Find..."
```
dubiousdavid: This should probably be customizable.

src/layout/TextHighlight.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+import * as F from 'futil-js'
+
+export default ({ pattern, text, Wrap = 'i' }) =>
+  pattern
+    ? F.highlight('<>', '<>', pattern, text)
+        .split('<>')
+        .map((x, i) => (i % 2 ? <Wrap key={i}>{x}</Wrap> : x))
```
sadasant: Is this wrapping in italics every other highlighted element? If so, why?

src/queryBuilder/FilterContents.js
```diff
@@ -0,0 +1,81 @@
...
+                  _.get(`${node.field}.typeOptions`, fields)
+                ),
+              ]
+            )}
+          </select>
+        </div>
+      )}
+      {node.type && (
+        <div
+          style={{
```
sadasant: This object can definitely be declared outside to avoid having to re-generate it every time. Probably in the `styles` file you have.

src/queryBuilder/QueryBuilder.js
```diff
@@ -0,0 +1,138 @@
...
+    .toString(36)
+    .substring(7)
+
+let { encode } = F.encoder('/') // todo get from client
+
+let blankNode = () => ({ key: randomString() })
+let replaceOn = (list, from, to) => list.splice(list.indexOf(from), 1, to)
+
+let ContextureClientBridge = (Types, Tree) => {
+  // not sure why reassignment is needed - its 2am and that makes it work... fix tomorrow!
```
sadasant: Is this comment still relevant?

src/queryBuilder/QueryBuilder.js
```diff
@@ -0,0 +1,138 @@
...
+  }))
+  return {
+    getNode: Tree.getNode,
+    add: tree => Tree.add(tree.path, observable(blankNode())),
+    remove: (tree, node) => Tree.remove(node.path),
+    join: (tree, join) => Tree.mutate(tree.path, { join }),
+    mutate: Tree.mutate,
+    typeChange(node, type) {
+      let path = node.path
+      let parentPath = _.dropRight(1, node.path)
```
sadasant: This line has `node.path` but the previous line has `path = node.path`. Not sure what's better, if changing this specific line to use `path` or changing all the other lines to use `node.path` and removing `path = node.path` since it doesn't seem to add much value.

src/queryBuilder/QueryBuilder.js
```diff
@@ -0,0 +1,138 @@
...
+      let tree = Tree.getNode(parentPath)
+      let index = tree.children.indexOf(node)
+      Tree.remove(path)
+      Tree.add(
+        parentPath,
+        observable({
+          key: node.key,
+          type,
+          field: node.field,
```
sadasant: Maybe what we can do is to decompose `key, field, path` from node and use them instead 🤔Whatever seems better (related to my previous comment).

src/exampleTypes/Query.js
```diff
@@ -0,0 +1,21 @@
...
+  observer(({ tree, node, ...props }) => (
+    <input
+      className="contexture-search-box"
+      value={node.query || ''}
+      onChange={e =>
+        tree.mutate(node.path, {
+          query: e.target.value,
+        })
+      }
+      placeholder="Search"
```
dubiousdavid: Same here.

src/queryBuilder/Rule.js
```diff
@@ -1,23 +1,21 @@
...
-    isDragging,
-  } = args
-  return connectDragSource(
+let Rule = ({
+  state,
+  node,
+  tree,
+  root,
+  connectDragSource,
+  // connectDragPreview,
```
sadasant: Is this still relevant?

src/utils/mobx-react-utils.js
```diff
@@ -35,5 +34,9 @@ export let lenservable = x => {
   return s
 }
 // let toggle = (val, lens, unset=undefined) => {
-//   f.set(f.view(lens) === val ? unset : val, lens)
+//   F.set(F.view(lens) === val ? unset : val, lens)
 // }
```
sadasant: Is this still relevant?

src/utils/mobx-react-utils.js
```diff
@@ -35,5 +34,9 @@ export let lenservable = x => {
...
+
+export let withStateLens = state => inject(() => F.lensOf(observable(state)))
+
+export let partial = _.curry((x, y) => inject(() => x)(y))
```
sadasant: Neat! 🍛💉

src/utils/schema.js
```diff
@@ -0,0 +1,49 @@
...
+      let typeDefault = F.alias(type, {
+        string: 'query',
+        text: 'facet',
+        long: 'number',
+        float: 'number',
+        double: 'number',
+      })
+      return {
+        typeDefault,
+        // TODO: exists, bool, geo, text? //date auto
```
sadasant: Is this still relevant?

stories/imdb/stories/filterList.js
```diff
@@ -0,0 +1,182 @@
...
+    _.update(
+      'movies.fields',
+      _.flow(
+        _.merge(_, {
+          released: {
+            label: 'Release Date',
+          },
+          // ...flagFields({
+          //   isCommon: ['plot', 'title'],
+          // }),
```
sadasant: Are these comments still relevant?

stories/queryBuilder/examples.js
```diff
@@ -0,0 +1,151 @@
...
+              join: 'or',
+              children: [
+                Node('query', 'filter 2a'),
+                Node('query', 'filter 2b'),
+                {
+                  key: 'group2',
+                  join: 'and',
+                  children: [
+                    Node('facet', 'filter 4a'),
+                    Node('query', 'filter 4b'),
```
sadasant: These examples are really pretty.

src/FilterAdder.js
```diff
@@ -0,0 +1,32 @@
...
+      )
+    }
+    return (
+      <Picker
+        options={options}
+        onChange={field => {
+          tree.add(path, {
+            key: _.uniqueId('add'),
+            field,
+            type: fields[field].typeDefault,
```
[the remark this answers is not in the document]
daedalus28: Nope, because the field itself comes from the same field list as generated on line 12. Filter adder needs fields to have `field` and `typeDefault` or it's not going to work, which you can see in the storybook. There might be a better way to document it - definitely open to suggestions

src/FilterList.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import * as F from 'futil-js'
+import _ from 'lodash/fp'
+import { observer } from 'mobx-react'
+import { Dynamic, SpacedList } from './layout'
+import InjectTreeNode from './utils/injectTreeNode'
+
+export let FieldLabel = InjectTreeNode(
+  observer(({ node: { field } = {}, fields }) => (
+    <b>{_.get(`${field}.label`, fields) || F.autoLabel(field)}</b>
```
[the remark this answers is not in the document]
daedalus28: Thanks for the reminder - this isn't actually supposed to still be there, fields should always have a label property thanks to `applyDefaults` from the schema util file

src/exampleTypes/Date.js
```diff
@@ -0,0 +1,26 @@
...
+import injectTreeNode from '../utils/injectTreeNode'
+
+export default injectTreeNode(
+  observer(({ tree, node, ...props }) => (
+    <Flex {...props}>
+      <input
+        className="contexture-search-box"
+        type="date"
+        value={node.from || ''}
+        onChange={e => tree.mutate(node.path, { from: e.target.value })}
```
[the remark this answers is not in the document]
daedalus28: We have lens utils already which we should probably use with tree lens that abstracts mutate

src/exampleTypes/DateHistogram.js
```diff
@@ -0,0 +1,38 @@
...
+import { Flex } from '../layout/Flex'
+import injectTreeNode from '../utils/injectTreeNode'
+
+export default injectTreeNode(
+  observer(
+    ({
+      node,
+      format = _.identity,
+      height = 100,
+      background = () => '#ccc',
```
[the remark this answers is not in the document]
daedalus28: This is an extension point so you can customize the colors of the bars. It defaults to always grey but you can make it contextual based on data.

src/exampleTypes/Facet.js
```diff
@@ -0,0 +1,53 @@
...
+import _ from 'lodash/fp'
+import * as F from 'futil-js'
+import { observer } from 'mobx-react'
+import { exampleTypes } from 'contexture-client'
+import { Flex } from '../layout/Flex'
+import injectTreeNode from '../utils/injectTreeNode'
+
+// For futil-js
+let toggleElement = (check, val, arr = []) =>
+  (check ? _.pull : F.push)(val, arr)
```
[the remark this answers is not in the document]
daedalus28: In this case order doesn't matter, facet iterates over context options

src/exampleTypes/Facet.js
daedalus28: That may be true, but these are exampleTypes. While we intend to use them in production, we should wait until there's a real use case for overriding the placeholder since every instance of facet currently has this placeholder in our main application. Example types will eventually be split into their own repo, so their API shouldn't be considered as first class as the core contexture-react components. Also, that's something we can easily add in a minor release but would have to release a major revision to remove it.

src/exampleTypes/Number.js
```diff
@@ -0,0 +1,26 @@
...
+import { exampleTypes } from 'contexture-client'
+import injectTreeNode from '../utils/injectTreeNode'
+
+export default injectTreeNode(
+  observer(({ tree, node, ...props }) => (
+    <Flex {...props}>
+      <input
+        className="contexture-search-box"
+        type="number"
+        value={node.min || ''}
```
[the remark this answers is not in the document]
daedalus28: That was added because if it's null, react thinks it's uncontrolled: https://github.com/smartprocure/contexture-react/pull/28/commits/d818577621cd9733aac6b5742c085325f2fd1b82

`null` and `undefiined` didn't work, and 0 is not correct.

src/exampleTypes/Query.js
daedalus28: Same reply as above

src/exampleTypes/ResultPager.js
daedalus28: Both are provided by contexture-client. That's a great example on why I want to pull example types out to their own repo instead of spread out across all the different packages.

src/exampleTypes/ResultPager.js
daedalus28: This is something to investigate separately and not as this PR. Worst case scenario here, this will cause the react reconciliation algorithm diff a single character of `>` on rerender.

src/exampleTypes/Styles.js
daedalus28: It is, but it's here to show the difference in these example styles just being about changing opacity. This file is pretty much throwaway code as a demo and isn't really intended to be used anywhere. In fact, this should probably get refactored to component-based overrides like the newer example types

dubiousdavid:
auto-table-sort
This feels like it could use some more polish. For example, `cursor: pointer` and some kind of hover style. Also, "Remove Column" changes the view from table to JSON in the IMDB example.

src/layout/Awaiter.js
daedalus28: Render props are the new hotness  😄 
https://reactjs.org/docs/render-props.html#using-props-other-than-render
https://medium.com/dailyjs/reacts-%EF%B8%8F-new-context-api-70c9fe01596b

src/layout/Flex.js
daedalus28: The one in marketplace is dependent on bootstrap and is why I didn't use it. I didn't need the other props yet so that's why it's marked as a TODO.

src/layout/Pickers.js
daedalus28: We don't know that's the case, and if so, we don't know what the performance impact will be. That's something to follow up on afterwards and shouldn't block 1.0

src/layout/Popover.js
daedalus28: Good point, contexture-react was started before we adopted `F` over `f`. There are probably a bunch of places that do this in the repo

src/layout/TextHighlight.js
daedalus28: Since we're using the same token for start and end and splitting on it, every even element in the matches would have been wrapped in the token and thus should be highlighted. It's also more efficient than writing a regex.

src/queryBuilder/FilterContents.js
daedalus28: Again, let's wait until we have confirmation that this will meaningfully impact performance and isn't optimizing microseconds like counting backwards on a for loop. I'd rather keep this as close to the usage as possible, especially since we're so early on in these components and aren't sure if these styles will stay as they are or be radically changed soon.

src/queryBuilder/QueryBuilder.js
daedalus28: Yes 😞

daedalus28: It is a mystery because it works in the unit test:

https://github.com/smartprocure/contexture-client/blob/master/test/index.js#L631-L641

and as you can see it is doing an extendOn which mutates: https://github.com/smartprocure/contexture-client/blob/master/src/index.js#L112-L116

src/queryBuilder/Rule.js
daedalus28: Nope, that's from when I was still experimenting with the react dnd api :D

src/utils/mobx-react-utils.js
daedalus28: nope, not used anymore

src/utils/schema.js
daedalus28: Yup, we are still missing those

src/queryBuilder/QueryBuilder.js
```diff
@@ -0,0 +1,138 @@
...
+      targetTree.children.splice(index, 0, node)
+    },
+    indent(tree, node, skipDefaultNode) {
+      // Reactors:
+      //   OR -> And, nothing
+      //   AND -> OR, others if has value
+      //   to/from NOT, others if has value
+      if (!tree) {
+        // Indent in place, only needs to happen at root and only works if node is a group
+        let key = randomString()
```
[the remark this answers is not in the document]
daedalus28: keys just need to be unique within their group. It's extremely unlikely to have a collision, and there is a plan for an autoKey which will generate keys much more intelligently in contexture-client

daedalus28:
@dubiousdavid yup the whole things needs way more polish, the idea is that we can 

1. override those as consumers of this lib and
2. have other developers besides me contribute to it 😃

src/exampleTypes/ResultTable.js
```diff
@@ -0,0 +1,109 @@
...
+    <th>
+      <a onClick={F.flip(popover)}>
+        {label}{' '}
+        {field === node.sortField && (node.sortDir === 'asc' ? '^' : 'v')}
+      </a>
+      <Popover isOpen={popover}>
+        <div style={{ textAlign: 'left' }}>
+          <div>
+            <a onClick={() => mutate({ sortField: field, sortDir: 'asc' })}>
+              ^ Sort Ascending
```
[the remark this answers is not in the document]
daedalus28: The plan is to allow these to be completely overridden with a react component

src/layout/Awaiter.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+import { observer } from 'mobx-react'
+
+export default observer(({ promise, children }) =>
+  promise.case({
+    pending: () => <div>Loading...</div>,
```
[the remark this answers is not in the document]
daedalus28: The plan is to allow Awaiter to take pending and error components as well - however from an internationalization standpoint there's a fair amount in here that would need work (e.g. everywhere we show and/or, etc)x

src/queryBuilder/QueryBuilder.js
```diff
@@ -0,0 +1,138 @@
...
+    ({ tree: iTree, types: iTypes }, { types = iTypes, tree = iTree }) => ({
+      types,
+      state: observable({
+        adding: false,
+        ...ContextureClientBridge(types, tree),
+      }),
+    }),
+    ({ state, path, fields, types = {} }) => (
+      <Provider fields={fields} types={types}>
+        <div style={{ background }}>
```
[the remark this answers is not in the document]
daedalus28: For the background, absolutely. This control can't have a transparent background because there are a few places where it uses divs to cover lines (haven't found a better solution 😅), so this needs to be able to be passed in

src/queryBuilder/QueryBuilder.js
sadasant: This is fixed by: https://github.com/smartprocure/contexture-client/pull/48

PR #32, opened 2018-04-24
[WIP] Feedback after usage

sadasant:
WIP WIP WIP

### DIFF
--- a/src/exampleTypes/Facet.js
+++ b/src/exampleTypes/Facet.js
@@ -11,7 +11,7 @@ let toggleElement = (check, val, arr = []) =>
   (check ? _.pull : F.push)(val, arr)
 
 export default injectTreeNode(
-  observer(({ tree, node, hide = {}, TextInput = 'input' }) => (
+  observer(({ tree, node, values, hide = {}, TextInput = 'input' }) => (
     <div>
       {!hide.facetFilter && (
         <TextInput
@@ -23,7 +23,7 @@ export default injectTreeNode(
         />
       )}
       {_.map(({ name, count }) => {
-        let checked = _.includes(name, node.values)
+        let checked = _.includes(name, node.values || values)
         return (
           <Flex
             key={name}
@@ -33,7 +33,7 @@ export default injectTreeNode(
               type="checkbox"
               onChange={() => {
                 tree.mutate(node.path, {
-                  values: toggleElement(checked, name, node.values),
+                  values: toggleElement(checked, name, node.values || values),
                 })
               }}
               checked={checked}
--- a/src/exampleTypes/Facet.js
+++ b/src/exampleTypes/Facet.js
@@ -11,7 +11,7 @@ let toggleElement = (check, val, arr = []) =>
   (check ? _.pull : F.push)(val, arr)
 
 export default injectTreeNode(
-  observer(({ tree, node, values, hide = {}, TextInput = 'input' }) => (
+  observer(({ tree, node, hide = {}, TextInput = 'input' }) => (
     <div>
       {!hide.facetFilter && (
         <TextInput
@@ -23,7 +23,7 @@ export default injectTreeNode(
         />
       )}
       {_.map(({ name, count }) => {
-        let checked = _.includes(name, node.values || values)
+        let checked = _.includes(name, node.values)
         return (
           <Flex
             key={name}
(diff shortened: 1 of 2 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

  The changelog has not been updated. Please update the changelog.
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>

  The version was not updated. Please update the version.
  </td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:book:</td>
      <td>

  No `duti:fix` npm script found. Not autofixing this PR.
  </td>
    </tr>
  </tbody>
</table>


<p align="right">
</p>

src/exampleTypes/Facet.js
```diff
@@ -11,7 +11,7 @@ let toggleElement = (check, val, arr = []) =>
   (check ? _.pull : F.push)(val, arr)
 
 export default injectTreeNode(
-  observer(({ tree, node, hide = {}, TextInput = 'input' }) => (
+  observer(({ tree, node, values, hide = {}, TextInput = 'input' }) => (
```
daedalus28: This shouldn't be necessary. If you want to set values from the Facet component you should be able to already as this inject tree node reads it off the reactor config from the facet example type

src/exampleTypes/Facet.js
```diff
@@ -33,13 +33,13 @@ export default injectTreeNode(
...
                 tree.mutate(node.path, {
-                  values: toggleElement(checked, name, node.values),
+                  values: toggleElement(checked, name, node.values || values),
                 })
               }}
               checked={checked}
             />
             <div style={{ flex: 2, paddingLeft: '5px', paddingRight: '5px' }}>
-              {name}
+              {`${name}`}
```
daedalus28: Why do this?

src/exampleTypes/ResultTable.js
```diff
@@ -86,14 +86,15 @@ export default InjectTreeNode(
...
-                        {display(getRecord(x)[field], getRecord(x))}
-                      </Cell>
-                    ),
-                    schema
-                  )}
+                  {_.map(({ field, display, Cell = 'td' }) => {
+                    let record = getRecord(x)
+                    let props = { key: field }
+                    // React warns you if you send record as a property to `td`
+                    if (Cell !== 'td') props.record = record
```
daedalus28: What's the reasoning on this? The display function is passed the record, why would you want it on an overridable component as well? Display can return a react component, so for conditional formatting you'd do it in there - the only reason to actually override the cell component is if a `td` doesn't work for some reason.

src/exampleTypes/ResultTable.js
```diff
@@ -23,7 +23,7 @@ let Header = withStateLens({ popover: false })(
     <th>
       <a onClick={F.flip(popover)}>
         {label}{' '}
-        {field === node.sortField && (node.sortDir === 'asc' ? '^' : 'v')}
+        {field === node.sortField && (node.sortDir === 'asc' ? '↑' : '↓')}
```
daedalus28: Neat :) The long term plan is to figure out how to expose this to be customized by consumers but this is an improvement

src/exampleTypes/Facet.js
daedalus28: In fact, this code won't actually even check the values on the facet because it doesn't call mutate with values. Was the goal to enable setting `values` with `path` usage instead of group where the component creates the node? If that doesn't work then the right approach is to extend injectTreeNode to handle that case as well if we think it's valuable (though I think you'd either have the components completely own setting up the nodes or you'd set them up in code, not have a weird mix where you set some node values from component props which is pretty confusing)

src/exampleTypes/ResultTable.js
daedalus28: The idea is to use `display` 99% of the time since in theory that can be reused by multiple components.

daedalus28:
Pretty sure all of these things have been fixed in other ways already and this PR is stale

PR #41, opened 2018-06-22
Renamed the Date component to DateComponent

### LINKED ISSUE #35: The `Date` component should have another name
The `Date` exampleType component should have another name. Currently, requiring this property alone overwrites the global object Date.

sadasant:
Fixes #35

### DIFF
--- a/src/exampleTypes/index.js
+++ b/src/exampleTypes/index.js
@@ -1,6 +1,6 @@
 import Facet from './Facet'
 import Number from './Number'
-import Date from './Date'
+import DateComponent from './DateComponent'
 import DateRangePicker from './DateRangePicker'
 import Query from './Query'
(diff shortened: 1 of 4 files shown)


src/exampleTypes/index.js
```diff
@@ -26,7 +26,7 @@ export default (
   let Components = {
     Facet: partial({ TextInput, Checkbox, RadioList }, Facet),
     Number: partial({ NumberInput }, Number),
-    Date: partial({ DateInput }, Date),
+    Date: partial({ DateInput }, DateComponent),
```
daedalus28: The key here needs to change too because there is no `Components.DateComponent,` on line 43.

daedalus28:
Closing as this is a trivial PR and is getting stale - the original ticket is still open though for when we are ok with the breaking change

PR #63, opened 2018-08-27
Conditionally show filter for ResultTable

dubiousdavid:
* Conditionally show filter for ResultTable.
* Add displayName to components.


dubiousdavid:
@daedalus28 Removed changes to `Number.js`.

PR #75, opened 2018-10-01
Highlighted Column

sadasant:
This PR adds a column when there are highlighted results. This column is called `Other Matches`, and shows a message if there were other fields that matched the given query, but that aren't visible due to the includes. The full values are shown in a table on a modal once any cell of the rows of this column is clicked.

**TODOs:** . 
- I'm not sure why CI is failing.

I think that would be all, besides general feedback :)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

<details>
  <summary>ResultTable.js</summary>
  <p><a href="https://github.com/smartprocure/contexture-react/blob/feature/highlightedColumn/src/exampleTypes/ResultTable.js" target="_blank">src/exampleTypes/ResultTable.js</a></p>
  
  <strong>Mixed spaces and tabs.</strong>
  <code>Line 72: <br /></code>
</details>
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>

  The changelog has not been updated. Please update the changelog.
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>

  The version was not updated. Please update the version.
  </td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


<p align="right">
</p>

src/exampleTypes/ResultTable.js
```diff
@@ -41,6 +41,84 @@ let popoverStyle = {
...
+                      alsoHighlighted
+                    )}
+                  </tr>
+                </thead>
+                <thead>
+                  <tr>
+                    {_.map(
+                      ([text]) => (
+                        <td>
+                          <TextHighlight
```
daedalus28: You shouldn't need to do this. The fields come from elasticsearch already highlighted, just render the HTML from the server directly

src/exampleTypes/ResultTable.js
```diff
@@ -41,6 +41,84 @@ let popoverStyle = {
...
+      visibleFields,
+      node,
+      highlight = _.result('slice.0.highlight', getResults(node)),
+      header,
+      Cell = header ? 'th' : 'td',
+      Table = 'table',
+      Modal = null,
+      viewModal,
+    }) => {
+      let alsoHighlighted = _.difference(_.keys(highlight), visibleFields)
```
daedalus28: We already solve this problem in the results type via `additionalFields` https://github.com/smartprocure/contexture-elasticsearch/blob/master/src/example-types/results.js#L73

src/exampleTypes/ResultTable.js
```diff
@@ -41,6 +41,84 @@ let popoverStyle = {
...
+      let getConnector = i => {
```
daedalus28: we use `toSentence` for this usually which is available on `underscore.string.fp`

daedalus28: https://github.com/stoeffel/underscore.string.fp#tosentencelastdelimiter-delimiter-array--string

If it doesn't support tree shaking (e.g. just picking that one method), we could add it to futil since I think it's the only method not in lodash already 🤔

src/exampleTypes/ResultTable.js
```diff
@@ -41,6 +41,84 @@ let popoverStyle = {
...
+        <Cell key="highlight">
+          {Modal && (
+            <Modal isOpen={viewModal}>
+              <h3>Other Matching Fields</h3>
+              <Table>
+                <thead>
+                  <tr>
+                    {_.map(
+                      x => (
+                        <th key={x}>{_.startCase(x)}</th>
```
daedalus28: We actually should check the schemas in case there are labels for those fields. Just because the fields aren't in the table currently, it doesn't mean that they're not still on the schemas

src/exampleTypes/ResultTable.js
sadasant: I can't render the HTML from the server directly since some of it has a whole HTML body and other structures :/ this is unsafe. We should clean it up first, then highlight.

src/exampleTypes/ResultTable.js
sadasant: Does it though? I'm testing and if the highlight object includes a property that is not in the includes, I don't see it in additionalFields

src/exampleTypes/ResultTable.js
sadasant: After talking with Sam about this:
> I'll just pass it along and see if it breaks, if it breaks I'll do the "remove all the other tags", and I'll do it in contexture elasticsearch

src/exampleTypes/ResultTable.js
sadasant: My chat with Sam:
> **Daniel:** so, what am I doing wrong? I'm removing "summary" from the includes, but I'm not getting "summary" on the additional fields
> **Sam:** Check the source of results, there might be a bug. Maybe we never tested it with includes? I mean that would be weird lol but maybe.

sadasant: https://github.com/smartprocure/contexture-elasticsearch/pull/95

src/exampleTypes/ResultTable.js
sadasant: Passing it is fine :)

src/exampleTypes/ResultTable.js
```diff
@@ -41,6 +41,67 @@ let popoverStyle = {
...
+                    )}
+                  </tr>
+                </thead>
+                <thead>
+                  <tr>
+                    {_.map(
+                      ({ value }) => (
+                        <td dangerouslySetInnerHTML={{ __html: value }} />
+                      ),
+                      additionalFields
```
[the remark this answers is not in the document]
sadasant: it’s not pretty at all. When I made this PR I left a couple of questions on the description. Essentially, this table is not getting the gray vest style somehow. Besides needing help there, I also need requirements 😅

sadasant:
I have been unable to reproduce the error reported by circleci :/

PR #78, opened 2018-10-04
add support for table layout in terms stats

dshishkov:
reference https://github.com/smartprocure/bid-search/issues/219


src/layout/Table.js
```diff
@@ -0,0 +1,108 @@
...
+                        details
+                      )(_.get(field, x), x)}
+                    </td>
+                  ),
+                  columns
+                )}
+              </tr>
+              {/* See if there is a details component to render for the column value when row expanded */}
+              {_.flow(
+                key => state.expandedRows.find(row => row.key === key),
```
sadasant: I wonder this would be better if we store the information as needed in the exampleRow. My intuition is that this might be easier to understand if it looks something like:

```javascript
{expandedRows[key] && <expandedRows[key].Component />}
```

So that the expandedRows is not an array but an object where the keys are the row.keys, and `expandedRow.component` is assigned like:
```javascript
expandedRows[recordKey] = {
  // ... the other props, then
  Component: () => _.get('component', details) && <tr>
  <td colSpan={columns.length}>
    {details.component(x[field], x)}
  </td>
</tr>
```

Sorry for the pseudocode! Let me know what you think!

stories/imdb/stories/filterList.js
```diff
@@ -108,7 +133,80 @@ export default () => (
...
+                      details: {
+                        expand: {
+                          display: x =>
+                            `Show top 50 based on meta score for ${x} ▼`,
+                        },
+                        collapse: {
+                          display: x =>
+                            `Hide top 50 based on meta score for ${x} ▲`,
+                        },
+                        component(x) {
```
sadasant: I love how extensible this table component is!

Little question: Would there be a better name for this property? I get why it's called component, but we usually use component in uppercase! There's no actual need on keeping consistency here, but makes me wonder if we could make the difference clearer (if not, disregard!).

src/layout/Table.js
```diff
@@ -0,0 +1,108 @@
...
+            </React.Fragment>
+          ),
+          data
+        )}
+      </tbody>
+    </table>
+  ))
+)
+
+Table.displayName = 'Table'
```
sadasant: I also wonder if the Results Table should use this Table component!

src/layout/Table.js
```diff
@@ -0,0 +1,108 @@
+import _ from 'lodash/fp'
+import * as F from 'futil-js'
+import React from 'react'
+import { observable } from 'mobx'
+import { observer, inject } from 'mobx-react'
+
+let Table = inject(() => ({
```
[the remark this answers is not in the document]
dshishkov: Thanks for the feedback! Yes, that name does actually sounds pretty nice considering the purpose of the component.

src/layout/Table.js
```diff
@@ -0,0 +1,108 @@
...
+              <tr {...x.rowAttrs} key={x[recordKey]}>
+                {F.mapIndexed(
+                  ({ field, display = x => x, details = {} }, i) => (
+                    <td
+                      key={`${field}${i}`}
+                      {...!_.isEmpty(details) && {
+                        style: {
+                          cursor: !_.isEmpty(details) ? 'pointer' : 'auto',
+                        },
+                        onClick() {
```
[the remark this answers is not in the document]
dshishkov: moved it into the inject, let me know if it makes it better to read/follow.

src/layout/Table.js
dshishkov: I tried that at first but since it is a dynamic key observable object, my observer was not picking up the changes, what do you think? I know we had something like expandObservable in the marketplace repo...

stories/imdb/stories/filterList.js
dshishkov: made it look for capital C

src/layout/Table.js
dshishkov: This component doesn't have knowledge of contexture trees but I guess the ResultsTable could use it for rendering, not sure what the scope/effort of such refactoring would be.

src/layout/Table.js
sadasant: Yes! so mobx 4 has a `set` method `import { set } from 'mobx'` which lets you alter an existing observable!

src/layout/Table.js
sadasant: Yeah, me neither. I guess what I _really_ wonder is if we have cases of result nodes being rendered with expandable tables? Are you aware of similar use cases?

dshishkov: So far no, but that sounds likely at some point.

src/layout/Table.js
dshishkov: I switched it to use an observable map (which seems that get and set in mobx do anyway when you use them). It cleaned up nicely.

dshishkov:
refactored the table view of terms stats into a separate component TermsStatsTable

src/layout/ExpandableTable.js
```diff
@@ -0,0 +1,105 @@
+import _ from 'lodash/fp'
+import * as F from 'futil-js'
+import React from 'react'
+import { observable } from 'mobx'
+import { observer, inject } from 'mobx-react'
+
+let ExpandableTable = inject(() => {
+  let state = {
+    expanded: observable(new Map()),
```
dubiousdavid: Are you using `Map` because the keys aren't strings?

stories/imdb/stories/filterList.js
```diff
@@ -108,7 +133,79 @@ export default () => (
                   path={['searchRoot', 'releases']}
                   format={formatYear}
                 />
-                <TermsStats path={['searchRoot', 'genreScores']} />
+                <TermsStatsTable
+                  path={['searchRoot', 'genreScores']}
+                  tableAttrs={{ style: { margin: 'auto' } }}
+                  columns={[
```
dubiousdavid: Is there any way we can accomplish this with child components?

src/layout/ExpandableTable.js
dshishkov: The main reason was because keys are dynamic for the each row that is expanded and the observer re-renders based on keys being added/removed from the expanded observable.

expanded: {
"KeyValue1": {},
"KeyValue5": {}
}

stories/imdb/stories/filterList.js
dshishkov: Do you mean so that we can wrap ExpandableTable within a TermsStats parent component? We would have to somehow pass the node from the path of the TermStats component as a prop to ExpandableTable. I am open to suggestions.

dshishkov: Something like this?

https://github.com/smartprocure/contexture-react/blob/feature/AddTableLayoutForTermsStatsType/stories/imdb/stories/filterList.js#L144-L178

dubiousdavid: Yeah I think that looks pretty good.

src/layout/ExpandableTable.js
```diff
@@ -4,7 +4,10 @@ import React from 'react'
 import { observable } from 'mobx'
 import { observer, inject } from 'mobx-react'
 
-let ExpandableTable = inject(() => {
+export let Column = _.identity
+Column.displayName = 'Column'
+
+let ExpandableTable = inject((stores, props) => {
```
dubiousdavid: This code could benefit from being slip into at least two, but probably even more, parts. Definitely the inject and the component parts.

stories/imdb/stories/filterList.js
```diff
@@ -136,76 +138,58 @@ export default () => (
...
+                    expand={{
+                      display: x =>
+                        `Show top 50 based on meta score for ${x} ▼`,
+                    }}
+                    collapse={{
+                      display: x =>
+                        `Hide top 50 based on meta score for ${x} ▲`,
+                    }}
+                  >
+                    {x => (
```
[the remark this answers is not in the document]
dshishkov: x is the value of the field for the current row/record, the one that needs to be acted on on expand.

src/layout/ExpandableTable.js
dshishkov: I split it up some, is it easier to follow this way?

src/layout/ExpandableTable.js
```diff
@@ -0,0 +1,134 @@
...
+        x => (
+          <React.Fragment key={x[recordKey]}>
+            <tr {...x.rowAttrs} key={x[recordKey]}>
+              {F.mapIndexed(
+                ({ field, display = x => x, details = {} }, i) => (
+                  <td
+                    key={`${field}${i}`}
+                    {...!_.isEmpty(details) && {
+                      style: {
+                        cursor: !_.isEmpty(details) ? 'pointer' : 'auto',
```
daedalus28: this _.isEmpty conditional will never be different since line 59 checks it too

PR #95, opened 2018-11-08
Fixes to CheckableResultsTable

sadasant:
Fixed unnecessary re-renderings on the CheckableResultsTable component.

### DIFF
--- a/src/exampleTypes/CheckableResultTable.js
+++ b/src/exampleTypes/CheckableResultTable.js
@@ -4,6 +4,7 @@ import { observer } from 'mobx-react'
 import F from 'futil-js'
 import InjectTreeNode from '../utils/injectTreeNode'
 import { getResults, getRecord } from '../utils/schema'
+import { len } from '../utils/futil'
 
 // Extends ResultTable with a checkbox column
 // Writes to a lens called `selected`, using getValue to map the selected record to a value.
@@ -12,7 +13,7 @@ let CheckableResultTable = InjectTreeNode(
   observer(
     ({ node, fields, selected, getValue, Checkbox, ResultTable, ...props }) => {
       let results = getResults(node).slice()
-      let allChecked = results.length === F.view(selected).length
+      let allChecked = len(results) === len(F.view(selected))
       let checkAll = F.sets(
         allChecked
           ? []
--- a/src/exampleTypes/CheckableResultTable.js
+++ b/src/exampleTypes/CheckableResultTable.js
@@ -3,12 +3,12 @@ import _ from 'lodash/fp'
 import { observer } from 'mobx-react'
 import F from 'futil-js'
 import { getResults, getRecord } from '../utils/schema'
-import { len } from '../utils/futil'
+import { length } from '../utils/futil'
 import InjectTreeNode from '../utils/injectTreeNode'
 
 let Label = observer(({ node, Checkbox, selected, getValue }) => {
   let results = getResults(node).slice()
-  let allChecked = len(results) === len(F.view(selected))
+  let allChecked = length(results) === length(F.view(selected))
   let checkAll = F.sets(
     allChecked
       ? []
--- a/src/exampleTypes/CheckableResultTable.js
+++ b/src/exampleTypes/CheckableResultTable.js
@@ -2,51 +2,50 @@ import React from 'react'
 import _ from 'lodash/fp'
 import { observer } from 'mobx-react'
 import F from 'futil-js'
-import InjectTreeNode from '../utils/injectTreeNode'
 import { getResults, getRecord } from '../utils/schema'
 import { len } from '../utils/futil'
+import InjectTreeNode from '../utils/injectTreeNode'
+
+let Label = observer(({ node, Checkbox, selected, getValue }) => {
+  let results = getResults(node).slice()
+  let allChecked = len(results) === len(F.view(selected))
+  let checkAll = F.sets(
+    allChecked
+      ? []
+      : _.map(
+          _.flow(
+            getRecord,
+            _.iteratee(getValue)
+          ),
+          results
+        ),
+    selected
+  )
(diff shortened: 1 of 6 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


<p align="right">
</p>

src/layout/Pickers.js
```diff
@@ -52,7 +52,31 @@ export let ModalPicker = withStateLens({ isOpen: false })(
...
         </Modal>
-        <Button onClick={F.on(isOpen)}>{label}</Button>
+        <Button onClick={F.on(isOpen)}>
+          {// React warns when the label is a component with children that don't
+          // have unique keys.  This is how we can force it. It's unlikely that
+          // we could be able to solve it differently, since the definition of
+          // uniqueness needs to happen at the rendering of the parent, and the
+          // label is rendered as a given property on the instatiation of our
+          // pickers (or their parent components).
+          typeof _.get('$$typeof', label) === 'symbol'
```
dubiousdavid: Can we use `_.isSymbol` here?

dubiousdavid: Also, if label is a symbol we treat it like an object? I'm confused what this is doing.

src/exampleTypes/CheckableResultTable.js
```diff
@@ -2,50 +2,49 @@ import React from 'react'
...
-      )
-    }
+let CheckableResultTable = observer(
+  ({ fields, selected, getValue, Checkbox, ResultTable, ...props }) => (
+    <ResultTable
+      fields={{
+        _checkbox: {
+          hideMenu: true,
+          label: '',
+          LabelComponent: getLabelComponent({ Checkbox, selected, getValue }),
```
daedalus28: You wouldn't need this additional API if you were just able to pass node to the label, which you could do if you leverage injectTreeNode on on CheckableResultTable like it was before. Also - I thought we were going to hold off on performance related things that complicate the code until we did profiling since we don't know objectively how much of a difference it makes. Did you start profiling stuff?

src/layout/Pickers.js
daedalus28: This code is very cryptic - and if the issue can be solved by just passing unique keys from the calling location, that makes this a user error and we should just add keys there. The relevant part is probably here: https://github.com/smartprocure/contexture-react/blob/master/src/themes/greyVest.js#L554-L557 However, we don't see this error on the contexture-imdb grey vest story, so it could be somewhere else.

src/utils/futil.js
```diff
@@ -12,3 +12,6 @@ export let FlattenTreeLeaves = Tree =>
   )
 export let PlainObjectTree = F.tree(onlyWhen(_.isPlainObject))
 export let flattenPlainObject = F.whenExists(FlattenTreeLeaves(PlainObjectTree))
+
+// Array
+export let len = _.get('length')
```
daedalus28: This is very close to `lens` which would be confusing if it was in the futil API. Ramda has a `length` function so there's precedent there if we really want to add this.

src/layout/Pickers.js
sadasant: I tried to comment why this code seemed like a good idea, but here's another take:
The `label` used by the Picker can be either a string or anything that could be passed to React. The way we're generating these components is very functional, so _I strongly believe_ React gets confused and thinks that the children of the `label` must have unique IDs (which doesn't always happen, specially doesn't happen when children are just written as HTML elements one below the other. It's common to see it when iterating an array).

Now, the alternative fix (which I pushed to this same branch a moment ago) is to just add these keys into `greyVest`, where it's needed. However: **it was considerably hard to track**, end users won't understand the problem, this shouldn't normally happen if you're adding elements within a `span` (while not iterating an array, for example). This is why this _symbol_ code seemed useful to me, it automatically fixes the unexpected necessity of a unique `key` property where it's not normally needed.

src/exampleTypes/CheckableResultTable.js
sadasant: I did this because the component was re-rendering too many times in my slow network, so it appeared as it was never going to render.

src/utils/futil.js
sadasant: Yes. Just let me know if I should put it somewhere else.

src/exampleTypes/CheckableResultTable.js
sadasant: This additional API is just to pass `Checkbox`, `selected` and `getValue`, not the node. Would `injectTreeNode` help to pass `Checkbox`, `selected` and `getValue`?

src/themes/greyVest.js
```diff
@@ -551,8 +551,10 @@ export let Adder = ModalFilterAdder({
         alignItems: 'center',
       }}
     >
-      Add Custom Filter
-      <i className="material-icons" style={{ opacity: 0.4 }}>
+      <div key={_.uniqueId('add-custom-filter')}>
```
dubiousdavid: This should be a span.

src/exampleTypes/CheckableResultTable.js
daedalus28: If you had node in scope, you could just have a regular component on label instead of needing to introduce labelComponent

sadasant: @daedalus28 not really, if you pass that regular component, it will be rendered at the point of declaration instead of being rendered as part of the table header, so it will still re-render the whole table.

PR #97, opened 2018-11-13
Nested field picker

daedalus28:
* Added nested support to the field picker. To nest options, add a `path` prop on fields before passing along to an Adder (which is also used by ResultsTable)
* Fixed Modal max-width to be fit-content instead of arbitrarily 500px
* Allow overriding the Picker in ModalFilterAdder

screen shot 2018-11-12 at 7 04 53 pm

### DIFF
--- a/src/layout/NestedPicker.js
+++ b/src/layout/NestedPicker.js
@@ -56,21 +56,21 @@ let PanelTreePicker = inject((store, { onChange, options }) => {
   let x = {
     state: observable({ selected: [] }),
     nestedOptions: toNested(options),
-    select: level => (key, field) => {
+    selectAtLevel: _.curry((level, key, field) => {
       if (isField(field)) onChange(field.value)
       else x.state.selected.splice(level, x.state.selected.length - level, key)
-    },
+    }),
   }
   return x
 })(
-  observer(({ select, state, nestedOptions, Item }) => (
+  observer(({ selectAtLevel, state, nestedOptions, Item }) => (
     <div
       className="panel-tree-picker"
       style={{ display: 'inline-flex', width: '100%', overflow: 'scroll' }}
     >
       <Section
         options={nestedOptions}
-        onClick={select(0)}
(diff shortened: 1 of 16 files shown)


src/layout/NestedPicker.js
```diff
@@ -0,0 +1,120 @@
...
+
+let toNested = _.flow(
+  _.map(x => _.defaults({ path: x.value }, x)),
+  unflattenObjectBy('path')
+)
+let PanelTreePicker = inject((store, { onChange, options }) => {
+  let x = {
+    state: observable({ selected: [] }),
+    nestedOptions: toNested(options),
+    select: level => (key, field) => {
```
sadasant: Would it be better to pass `level` as a property of an object? so it would be more clear when you call it `select({ level: 0 })`. Or perhaps rename `select` to be something like `selectLevel`.

src/layout/NestedPicker.js
```diff
@@ -0,0 +1,120 @@
...
+        Highlight={Highlight}
+        Item={Item}
+      />
+    ) : (
+      <PanelTreePicker options={options} onChange={onChange} Item={Item} />
+    )}
+  </div>
+)
+
+export default wrapPicker(NestedPicker)
```
sadasant: I think wrapPicker should be closer to this line. Also, if we're separating the definition of observer, can we define a name for all of these components?

daedalus28: I'll move wrapPicker - but we don't need displayName due to how it's declared. Babel infers the function name from the declaration

src/layout/NestedPicker.js
daedalus28: The represents selecting a field at a certain level, but agree that the nature of the currying isn't as clear as it could be 🤔

sadasant: It's not the currying, it's she 'select', since we're also talking about a nested `picker`. Select in my head is about picking an element from a list, not between levels.

PR #136, opened 2019-01-15
New move column algorithm implementation.

### LINKED ISSUE #121: Reordering columns in Result Table Doesn't work when omitting fields from include
If you have fields in the includes array which are omitted from the `fields` prop (to hide them from the table), the reordering logic won't work right unless all omitted fields are always at the end of the includes array. Here's the RCA from the ticket where we identified the issue in our product using this library:

> Incorrect/non functional reording behavior is caused by omitting fields that are present in the includes array. The indexes are off because the array index `i` of the `visibleFields` referenced here https://github.com/smartprocure/contexture-react/blob/master/src/exampleTypes/ResultTable.js#L150 isn't actually the index of the field in the actual includes array, it's the index of the includes array filtering out anything we're hiding. So for example, if includes is `[a, b, c, d]` and you omit `b`, the visible column index for `c` is going to be `1` even though it's really `2` on the actual `includes` array.
>
> This can be fixed a few different ways. We could:
> 
> - Patch contexture react so that when moving a column, we move it to be +/- 1 of the index whereever the occurence of the next visible item is (because if you have [a, b, c, d, e, f] and c, d, and e are all hidden, you have to know that moving b to the right actually means moving it to the end).
> - Just make sure that hidden fields are always at the end of the includes array.

geosp:
Fixes #121

### DIFF
--- a/src/exampleTypes/ResultTable.js
+++ b/src/exampleTypes/ResultTable.js
@@ -98,7 +98,7 @@ HeaderCellDefault.displayName = 'HeaderCellDefault'
 
 let Header = withStateLens({ popover: false, adding: false, filtering: false })(
   observer(({ // Local State
-    i, popover, adding, filtering, Modal, FieldPicker, ListGroupItem: Item, typeComponents, HeaderCell = HeaderCellDefault, field: fieldSchema, includes, addOptions, addFilter, tree, node, mutate, criteria, mapNodeToProps, fields, visibleFields, Icon }) => {
+    popover, adding, filtering, Modal, FieldPicker, ListGroupItem: Item, typeComponents, HeaderCell = HeaderCellDefault, field: fieldSchema, includes, addOptions, addFilter, tree, node, mutate, criteria, mapNodeToProps, fields, visibleFields, Icon }) => {
     // Components (providerable?) // Contextual
     let {
       disableFilter,
@@ -325,8 +325,8 @@ let ResultTable = InjectTreeNode(
         <thead>
           <tr>
             {F.mapIndexed(
-              (x, i) => (
-                <Header key={x.field} field={x} i={i} {...headerProps} />
+              x => (
+                <Header key={x.field} field={x} {...headerProps} />
               ),
               visibleFields
             )}
--- a/src/exampleTypes/ResultTable.js
+++ b/src/exampleTypes/ResultTable.js
@@ -133,7 +133,7 @@ let Header = withStateLens({ popover: false, adding: false, filtering: false })(
         includes
       )
       mutate({
-        include: F.moveIndex(_.head(indexes), _.last(indexes), [...includes]),
+        include: F.moveIndex(_.head(indexes), _.last(indexes), includes),
       })
     }
     let Label = label
--- a/src/exampleTypes/ResultTable.js
+++ b/src/exampleTypes/ResultTable.js
@@ -18,14 +18,24 @@ import {
 let getIncludes = (schema, node) =>
   F.when(_.isEmpty, _.map('field', schema))(node.include)
 
-const moveColumn = (mutate, computeNextIndex, field, visibleFields, includes) => {
+const moveColumn = (
+  mutate,
+  computeNextIndex,
+  field,
+  visibleFields,
+  includes
+) => {
   let visibleFieldIndex = _.findIndex({ field }, visibleFields)
   let nextField = _.flow(
     _.nth(computeNextIndex(visibleFieldIndex)),
     _.get('field')
   )(visibleFields)
   mutate({
-    include: F.moveIndex(_.indexOf(field, includes), _.indexOf(nextField, includes), includes),
+    include: F.moveIndex(
+      _.indexOf(field, includes),
+      _.indexOf(nextField, includes),
+      includes
+    ),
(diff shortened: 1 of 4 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

<details>
  <summary>ResultTable.js</summary>
  <p><a href="https://github.com/smartprocure/contexture-react/blob/feature/121_new_move_column_algorithm/src/exampleTypes/ResultTable.js" target="_blank">src/exampleTypes/ResultTable.js</a></p>
  
  <strong>'i' is defined but never used.</strong>
  <code>Line 101: i, popover, adding, filtering, Modal, FieldPicker, ListGroupItem: Item, typeComponents, HeaderCell = HeaderCellDefault, field: fieldSchema, includes, addOptions, addFilter, tree, node, mutate, criteria, mapNodeToProps, fields, visibleFields, Icon }) => {</code>
</details>
  </td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 024e9796b19055e00e00386ade036e391cef4ea2
</p>

src/exampleTypes/ResultTable.js
```diff
@@ -116,6 +116,18 @@ let Header = withStateLens({ popover: false, adding: false, filtering: false })(
       tree.mutate(filterNode.path, { paused: false })
       F.flip(filtering)()
     }
+    const moveColumn = (computeNewIndex, field, visibleFields) => {
+      let visibleFieldIndex = _.findIndex(x => x.field === field, visibleFields)
```
daedalus28: I think this can be 
```js
let visibleFieldIndex = _.findIndex({ field }, visibleFields)
```

src/exampleTypes/ResultTable.js
```diff
@@ -116,6 +116,18 @@ let Header = withStateLens({ popover: false, adding: false, filtering: false })(
...
+      let nextField = _.flow(
+        _.nth(computeNewIndex(visibleFieldIndex)),
+        _.get('field')
+      )(visibleFields)
+      let currentIndex = _.findIndex(x => x === field, [...includes])
+      let nextFieldIndex = _.findIndex(x => x === nextField, [...includes])
```
daedalus28: Since you're matching exact values, this would be cleaner with `_.indexOf`:

```js
let currentIndex = _.indexOf(field, [...includes])
let nextFieldIndex = _.indexOf(nextField, [...includes])
```

daedalus28: I'm also not sure we actually need to spread

src/exampleTypes/ResultTable.js
```diff
@@ -116,6 +116,18 @@ let Header = withStateLens({ popover: false, adding: false, filtering: false })(
       tree.mutate(filterNode.path, { paused: false })
       F.flip(filtering)()
     }
+    const moveColumn = (computeNewIndex, field, visibleFields) => {
```
daedalus28: For performance, it might be worth factoring this method out of the React component so it's not redeclared on each render. The only addition closure dependency it has seems to be `mutate`

stories/imdb/stories/greyVest.js
```diff
@@ -339,7 +340,10 @@ export default () => (
                 <div className="gv-box">
                   <ResultTable
                     path={['root', 'results']}
-                    fields={schemas[tree.tree.schema].fields}
+                    fields={_.omit(
+                      ['imdbId'],
```
daedalus28: we should validate this further with a test case that has at least two fields omitted to prove it works

src/exampleTypes/ResultTable.js
geosp: Fixed

src/exampleTypes/ResultTable.js
geosp: Fixed

stories/imdb/stories/greyVest.js
geosp: Fixed

src/exampleTypes/ResultTable.js
daedalus28: it's still defined inside of `Header` which means it'll get redeclared every render

stories/imdb/stories/greyVest.js
```diff
@@ -121,13 +121,15 @@ let tree = Contexture({
       key: 'results',
       type: 'results',
       include: [
+        'imdbId',
```
daedalus28: we should probably have the two omitted fields next to each other so we can see that it handles that

PR #137, opened 2019-01-15
Do not pass props indiscriminately to Section child

stellarhoof:
The `Section` component had an implicit interface assumption on its
`Item` prop (that it accepts `hasChildren` and `active` props). However,
when `Item` defaults to `div`, React complains about the extraneous
props.

### DIFF
--- a/src/layout/NestedPicker.js
+++ b/src/layout/NestedPicker.js
@@ -14,7 +14,7 @@ let unflattenObjectBy = _.curry((iteratee, x) =>
 let isField = x => x.typeDefault
 
 let FilteredSection = observer(
-  ({ options, onClick, highlight, Highlight, Item }) => (
+  ({ options, onClick, highlight, Highlight, Item = 'div' }) => (
     <div>
       {F.mapIndexed(
         (option, field) => (
@@ -28,20 +28,29 @@ let FilteredSection = observer(
   )
 )
 FilteredSection.displayName = 'FilteredSection'
-let Section = observer(({ options, onClick, selected, Item }) => (
+let Section = observer(({ options, onClick, selected, Item = 'div' }) => (
   <div>
     {F.mapIndexed(
-      (item, key) => (
-        <Item
-          key={key}
-          onClick={() => onClick(item.value || key, item)}
-          active={selected === key}
-          disabled={selected && selected !== key}
-          hasChildren={!isField(item)}
-        >
-          {isField(item) ? item.shortLabel || item.label : _.startCase(key)}
-        </Item>
-      ),
+      (item, key) =>
+        Item === 'div' ? (
+          <div
+            key={key}
+            onClick={() => onClick(item.value || key, item)}
+            disabled={selected && selected !== key}
+          >
+            {isField(item) ? item.shortLabel || item.label : _.startCase(key)}
+          </div>
+        ) : (
+          <Item
+            key={key}
+            onClick={() => onClick(item.value || key, item)}
+            disabled={selected && selected !== key}
+            active={selected === key}
+            hasChildren={!isField(item)}
+          >
+            {isField(item) ? item.shortLabel || item.label : _.startCase(key)}
+          </Item>
+        ),
       options
     )}
   </div>
@@ -98,7 +107,7 @@ let NestedPicker = ({
   filter,
   Input = 'input',
   Highlight = TextHighlight,
-  Item = 'div',
+  Item,
 }) => (
   <div>
     <Input {...F.domLens.value(filter)} placeholder="Enter filter keyword..." />
--- a/src/layout/NestedPicker.js
+++ b/src/layout/NestedPicker.js
@@ -13,8 +13,11 @@ let unflattenObjectBy = _.curry((iteratee, x) =>
 
 let isField = x => x.typeDefault
 
+const div = ({ children, onClick, disabled }) =>
+  <div onClick={onClick} disabled={disabled}>{children}</div>
+
 let FilteredSection = observer(
-  ({ options, onClick, highlight, Highlight, Item = 'div' }) => (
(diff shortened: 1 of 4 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Branch being merged does not follow [Git Flow](http://nvie.com/posts/a-successful-git-branching-model/)
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  Your PR has lint warnings. Please consider fixing these.
      
<details>
  <summary>NestedPicker.js</summary>
  <p><a href="https://github.com/smartprocure/contexture-react/blob/props_bugfix/src/layout/NestedPicker.js" target="_blank">src/layout/NestedPicker.js</a></p>
  
  <strong>Expected '===' and instead saw '=='.</strong>
  <code>Line 40: Item == 'div' ? {} : {</code>
</details>
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


Take a look at this commit to see what happened in detail: ff4f248e0ce6a50ea3b564523724404443d8707e


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against c9ece19e9a64b06ec6f0c9ab1608ee5819e45bbc
</p>

src/layout/NestedPicker.js
```diff
@@ -28,16 +28,20 @@ let FilteredSection = observer(
...
   <div>
     {F.mapIndexed(
       (item, key) => (
         <Item
           key={key}
           onClick={() => onClick(item.value || key, item)}
-          active={selected === key}
           disabled={selected && selected !== key}
-          hasChildren={!isField(item)}
+          {...(Item === 'div'
```
dubiousdavid: I think this is a little more straightforward if you just check if `Item` exists and if not ```
<div onClick="" key="">..</div>```. There is some small duplication, but the code is ultimately a little easier to read IMO.

stellarhoof: Is that what you had in mind? I personally like `const extraProps = ...` and then `<Item {...extraProps}>` more

stellarhoof: Sorry my force-pushing made this comment break since the commit hashes are not valid anymore. See https://github.com/smartprocure/contexture-react/pull/137/files#diff-9408d26625840440e038edcea38b6d24R34

src/layout/NestedPicker.js
```diff
@@ -28,20 +28,29 @@ let FilteredSection = observer(
   )
 )
 FilteredSection.displayName = 'FilteredSection'
-let Section = observer(({ options, onClick, selected, Item }) => (
+let Section = observer(({ options, onClick, selected, Item = 'div' }) => (
```
dubiousdavid: You should drop this and just check if `Item` is truthy then `Item` else `div`.

stellarhoof: I can't. React treats uppercase props specially and it complains if `Item` is ever undefined:
```
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports. Check the render method of `Section`.
```

dubiousdavid: Just go with your previous solution.

src/layout/NestedPicker.js
```diff
@@ -28,20 +28,29 @@ let FilteredSection = observer(
...
-          onClick={() => onClick(item.value || key, item)}
-          active={selected === key}
-          disabled={selected && selected !== key}
-          hasChildren={!isField(item)}
-        >
-          {isField(item) ? item.shortLabel || item.label : _.startCase(key)}
-        </Item>
-      ),
+      (item, key) =>
+        Item === 'div' ? (
```
daedalus28: Instead of doing this branching, a simpler approach is probably to just define a `DefaultItem` component that renders a div and white lists the props it passes through. This is also way more explicit about the assumptions contexture-react makes about the props it expects its components to take - special casing `div` isn't right, because that leaves it open to break for a `span` or something else. The idea is that we're saying "an `Item` component takes these props", with a default component that ignores `active` and `hasChildren` and passes the rest along to a `div`.

Also, if this is for a GreyVest scenario, I'd be surprised if the provided `Item` is actually just a div. GreyVest is the most robust theme we have - are you sure it's not using these?

daedalus28:
Also, don’t forget to update the package version and change log which are examples of things our CI automatically tells you about

src/layout/NestedPicker.js
```diff
@@ -13,6 +13,12 @@ let unflattenObjectBy = _.curry((iteratee, x) =>
 
 let isField = x => x.typeDefault
 
+const div = ({ children, onClick, disabled }) => (
```
dubiousdavid: You should call this `ItemDefault`.

stellarhoof: Changed. In that case heads up: https://github.com/smartprocure/contexture-react/blob/master/src/exampleTypes/ResultPager.js#L8, since I was following its example.

stellarhoof:
Who merges this?

PR #152, opened 2019-02-11
Feature/151

### LINKED ISSUE #151: Modal styling changes
Need a couple styling changes to the example types default modal to account for the fact that we get some pretty unruly content in there

* set a `px` max-width instead of `fit-content` since that's experimental and isn't working with the crazy nested tables that come through
* `overflow-x: scroll`
* change the current padding to a border so x-scrollable content scrolls within the visible whitespace instead of over it

doug-patterson:
Fixes #151 
Fixes https://github.com/smartprocure/bid-search/issues/1067

* styling changes to make the default modal better handle un-cooperative content

### DIFF
--- a/src/layout/Modal.js
+++ b/src/layout/Modal.js
@@ -20,7 +20,7 @@ let Modal = ({ isOpen, children, style = {} }) => (
           ...style,
         }}
         onClick={F.off(isOpen)}
-        className='default-modal-bg'
+        className="default-modal-bg"
       >
         <div
           style={{
@@ -33,7 +33,7 @@ let Modal = ({ isOpen, children, style = {} }) => (
             color: '#555',
           }}
           onClick={e => e.stopPropagation()}
-          className='default-modal-wrap'
+          className="default-modal-wrap"
         >
           {children}
         </div>
--- a/src/layout/Modal.js
+++ b/src/layout/Modal.js
@@ -24,10 +24,11 @@ let Modal = observer(
           style={{
             backgroundColor: '#fff',
             borderRadius: 5,
-            maxWidth: 'fit-content',
+            maxWidth: 640,
+            overflowX: 'scroll',
             // minHeight: 300,
             margin: '0 auto',
-            padding: 30,
+            border: '30px solid white',
             color: '#555',
           }}
           onClick={e => e.stopPropagation()}
--- a/src/layout/Modal.js
+++ b/src/layout/Modal.js
@@ -20,19 +20,20 @@ let Modal = ({ isOpen, children, style = {} }) => (
           ...style,
         }}
         onClick={F.off(isOpen)}
+        className='default-modal-bg'
       >
         <div
           style={{
             backgroundColor: '#fff',
             borderRadius: 5,
-            maxWidth: 640,
-            overflowX: 'scroll',
+            maxWidth: 'fit-content',
+            padding: 30,
             // minHeight: 300,
             margin: '0 auto',
-            border: '30px solid white',
             color: '#555',
           }}
           onClick={e => e.stopPropagation()}
+          className='default-modal-wrap'
         >
           {children}
         </div>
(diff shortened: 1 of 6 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 404b05472670035bc2408728dfc28d5d5a6bfbb9
</p>

src/layout/Modal.js
```diff
@@ -25,10 +25,11 @@ let Modal = ({ isOpen, children, style = {} }) => (
           style={{
             backgroundColor: '#fff',
             borderRadius: 5,
-            maxWidth: 'fit-content',
+            maxWidth: 640,
+            overflowX: 'scroll',
             // minHeight: 300,
             margin: '0 auto',
-            padding: 30,
+            border: '30px solid white',
```
doug-patterson: @daedalus28 we could also make the component take an `innerStyle` prop for this div and pass this through from grey vest

src/layout/Modal.js
```diff
@@ -25,10 +25,11 @@ let Modal = ({ isOpen, children, style = {} }) => (
           style={{
             backgroundColor: '#fff',
             borderRadius: 5,
-            maxWidth: 'fit-content',
+            maxWidth: 640,
+            overflowX: 'scroll',
             // minHeight: 300,
             margin: '0 auto',
-            padding: 30,
```
doug-patterson: changing from padding to border as this means the scrollable x overflow won't overlap the visible whitespace

src/layout/Modal.js
daedalus28: We should definitely do that. Also, the 30px border is super weird 🤔

daedalus28: And if that's too tricky, we can just put a class on it and override in grey vest with css - but the 30px should not be in the core modal code

doug-patterson:
@daedalus28 updated to add a couple classes to the default modal markup and then use those from the theme to style the modal. I've kept the 30px border as it's the best way I know of to get the desired effect (side scrolling inside of whitespace within a single element) -- to do it without that css hack would require nesting a second div (which would be fine if we prefer it that way, but I don't really see an issue with using the border for this purpose)

src/layout/Modal.js
```diff
@@ -20,18 +20,20 @@ let Modal = ({ isOpen, children, style = {} }) => (
...
         }}
         onClick={F.off(isOpen)}
+        className="default-modal-bg"
       >
         <div
           style={{
             backgroundColor: '#fff',
             borderRadius: 5,
             maxWidth: 'fit-content',
+            padding: 30,
```
daedalus28: let's just kill the default padding, border radius, color, etc that really belongs in grey vest (and so we don't need !important)

doug-patterson:
@daedalus28 updated

src/themes/greyVest.js
```diff
@@ -666,6 +666,14 @@ export let GVStyle = () => (
         padding: 15px;
         width: 400px !important;
       }
+
+      /* Modals */
+      .default-modal-wrap {
+        width: auto;
+        max-width: 640px!important;
```
daedalus28: Do we still need `!important`?

doug-patterson:
@daedalus28 updated

Issue #139, opened 2019-01-17
Tag Input State Persistence


daedalus28:
TagsInput loses it's state between renders, which means that the tag currently being typed resets to empty string. For some reason, inputs are getting rerendered in some of search pages more than they should (likely due to loading indicators we don't really need there), but this also prevents us from implementing the "backspace" functionality to restore a tag. The plan was to wait for react 16.8's hooks to implement functional state, but we might be able to nest components deeper or something to work around it now.

doug-patterson:
@daedalus28 just from looking at it it looks to me that the state object holding the current input needs to be injected into TagsQuery and then pushed down to TagsInput, which could be modified to use a passed in state object when found rather than creating its own on every render. That ought to fix it, but would complicate the relationship between TagsQuery and TagsInput in a way that for all I know is undesired. If that sounds like an OK plan, though, confirm here and I'll give it a shot.

daedalus28:
Even that won't work, because the issue is actually TagsQuery being rerendered, which is likely due to injectTreeNode wrapping it up for loading. The full fix here is probably to make TagsInput a legitimately stateful component using actual react component state, but in the mean time we can (and should) attempt to reduce the amount of rerendering that's happening on the TagsQuery component

PR #161, opened 2019-02-22
Preserve state between renders on TagsInput

stellarhoof:
Fixes https://github.com/smartprocure/contexture-react/issues/139

`TagsInput` was wrapped on an `inject` to use state, but `inject`'s purpose is to provide props, so the state got injected anew on every re-render. A more reusable solution is to have somethin g like `withState` for MobX, but this quickfix seems appropriate for now, since we'll probably move to React 16.8 with hooks later on.

### DIFF
--- a/src/layout/TagsInput.js
+++ b/src/layout/TagsInput.js
@@ -1,6 +1,7 @@
 import React from 'react'
 import _ from 'lodash/fp'
 import F from 'futil-js'
+import { withState } from 'recompose'
 import { observable } from 'mobx'
 import { observer, inject } from 'mobx-react'
 import { Flex } from './Flex'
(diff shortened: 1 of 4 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3f4339710a18ea8d688e38aad96ec0578e90ef46
</p>

stellarhoof:
When we merge this PR it'd be good to go over components using inject for their state and apply whatever solution we come up with here.

stellarhoof:
> The solution is just to declare the observable state outside the component

We won't be able to go with this as the state is then shared across all the instances of the component :( I'm reverting back to `withState` for now.

dubiousdavid:
Sorry for leading you down the rabbit-hole there :)

src/layout/TagsInput.js
```diff
@@ -27,13 +28,17 @@ let Tag = observer(({ value, removeTag, tagStyle, onClick }) => (
...
 Tag.displayName = 'Tag'
 
-let TagsInput = inject(() => ({
-  state: observable({
+// We're only using withState to preserve the state between renders, since
+// inject doesn't do that for us.
+let TagsInput = withState(
+  'state',
+  'setState',
+  observable({
```
dubiousdavid: Doesn't this need to be `() => observable({})`?

stellarhoof: That's one option yes, but they also accept a value (https://github.com/acdlite/recompose/blob/master/docs/API.md#withstate)

stellarhoof: I'm not sure I follow. On the code you linked, the state doesn't get overriden on re-renders since `this.state` is defined outside of `render()`.

dubiousdavid: Not sure how else to explain this other than objects are just references in Javascript. Seems like you're passing the same reference to every instance of the component, which would make them all connected. Could be wrong. This stuff is non-trivial to test locally.

stellarhoof: You're right. Also it works that way too in pretty much any language I've seen so it was just my brain fart. Fixed now. Also tested.

PR #175, opened 2019-02-27
 Fix for Popovers in ResultTables

sadasant:
Popovers were clashing with Modals (that use Portals).

### DIFF
--- a/src/exampleTypes/ResultTable.js
+++ b/src/exampleTypes/ResultTable.js
@@ -54,7 +54,7 @@ let HighlightedColumnHeader = observer(
       _.isEmpty
     )(results),
   }) =>
-    hasAdditionalFields ? (
+    hasAdditionalFields && node.showOtherMatches ? (
       <Cell key="additionalFields">Other Matches</Cell>
     ) : null
 )
@@ -283,15 +283,17 @@ let TableBody = observer(
                 ),
                 visibleFields
               )}
-              <HighlightedColumn
-                {...{
-                  node,
-                  additionalFields: _.result('additionalFields.slice', x),
-                  Modal,
-                  Table,
-                  schema,
-                }}
-              />
+              {node.showOtherMatches && (
+                <HighlightedColumn
+                  {...{
+                    node,
+                    additionalFields: _.result('additionalFields.slice', x),
+                    Modal,
+                    Table,
+                    schema,
+                  }}
+                />
+              )}
             </Row>
           ),
           getResults(node)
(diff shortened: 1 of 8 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against dc7f733ce10b4f7d490fa532e848162b28472633
</p>

src/exampleTypes/ResultTable.js
```diff
@@ -156,7 +156,14 @@ let Header = withStateLens({ popover: false, adding: false, filtering: false })(
           )}
           {hideMenu ? null : <Icon icon="TableColumnMenu" />}
         </span>
-        <Popover isOpen={popover} style={popoverStyle}>
+        <Popover
+          isOpen={x => {
+            // Only turn off the popover if adding is not true
+            if (!F.view(adding) && _.isBoolean(x)) return F.off(popover)()
```
doug-patterson: this fixes the issue (I tested) but I'm concerned about these features of it

* it makes it the case that the `popover` variable no longer controls whether the popover is open, which doesn't seem great
* it relies on the coincidence that the only situation in which we want to show the popover even though `popover` is false is the situation where we're adding a column.

Seems like it would be better to refactor so that `popover` continues to determine whether the popover is open. 

The code does work though so I'll leave this as a comment and not a request for changes -- I'd like to hear what others think, too

sadasant: the issue I see here is that `isOpen` is a function, which allows us to control in which situations it should or should not be open, which fits this change, in the following sense: On the ResultTable, the only situation in which we don't want the popover to accidentally close, is if the modal is open.

I'm not a fan of our current use of lenses. Lenses appear like properties that determine the current state of something, but can behave like event functions, such as onClick, or onChange. I find this unfortunately confusing.

doug-patterson: @sadasant you're more or less agreeing with me that it's just due to relatively coincidental facts about the current wiring of ResultTable that `adding: true` happens to be the other condition under which you want to show the popover. But that's an accidental fact about the `adding` state bool, which is supposed to indicate whether you're adding a field, and it also makes it the case that we can have `popover: false` and yet it be expected that the popover is open. Something should at least get renamed here, but better would be to wire it up so that the condition under which the popover is open is exactly `popover: true`. Is there something about the outside click handler stuff that prevents doing it in that clean way?

PR #200, opened 2019-03-21
Preserve geo location dropdown state

stellarhoof:
The `location` field returned from `loadHereOptions` was only being used by the Geo example type to store it in the node. However, it was slightly different from the label of the selected option, which was also a location. This PR changes that so we don't return `location` anymore and instead use the option's label as the `location` value for the node.

### DIFF
--- a/src/exampleTypes/Geo.js
+++ b/src/exampleTypes/Geo.js
@@ -63,6 +63,7 @@ let GeoComponent = injectTreeNode(
             <AutoComplete
               cacheOptions
               escapeClearsValue
+              defaultValue={node.selectedOption}
               placeholder={placeholder}
               noOptionsMessage={() => ''}
               styles={customStyles}
@@ -71,10 +72,11 @@ let GeoComponent = injectTreeNode(
                 const inputValue = newValue.replace(/[^a-zA-Z0-9\s]+/g, '')
                 return inputValue
               }}
-              onChange={async ({ value }) => {
-                let data = await GeoCodeLocation(value)
+              onChange={async selectedOption => {
+                let data = await GeoCodeLocation(selectedOption.value)
                 if (data && data.latitude && data.longitude && data.location) {
                   tree.mutate(node.path, {
+                    selectedOption,
                     latitude: data.latitude,
                     longitude: data.longitude,
                     location: data.location,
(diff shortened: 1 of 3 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 067384c51b7eaa8445cc7776cbea94ed8d29dd41
</p>

src/exampleTypes/Geo.js
```diff
@@ -71,10 +72,11 @@ let GeoComponent = injectTreeNode(
                 const inputValue = newValue.replace(/[^a-zA-Z0-9\s]+/g, '')
                 return inputValue
               }}
-              onChange={async ({ value }) => {
-                let data = await GeoCodeLocation(value)
+              onChange={async selectedOption => {
+                let data = await GeoCodeLocation(selectedOption.value)
                 if (data && data.latitude && data.longitude && data.location) {
                   tree.mutate(node.path, {
+                    selectedOption,
```
daedalus28: As a general rule, we should not add component props to contexture nodes. Contexture nodes do not necessarily map one to one with react components. In practice, they often do - but that's not guaranteed to be the case. 

Also, this would result in saving "label" in the database, which isn't right. Instead, we can just have the value be the first option that matches latitude and longitude - no need to persist this intermediate state

daedalus28:
As a reminder - per our discussion, the plan is to just put the label value on drop down change in the `location` value

PR #201, opened 2019-03-22
Overridable Loading component in injectTreeNode

stellarhoof:
This PR is also a proof of concept of how to theme contexture-react
using React's context mechanism, which MobX so kindly provides for us

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 1.43.9
+* Ability to override the Loading component in injectTreeNode
+
 # 1.43.8
 * Set overflow: auto on table's horizontal axis
 
--- a/src/utils/injectTreeNode.js
+++ b/src/utils/injectTreeNode.js
@@ -1,4 +1,6 @@
+import React from 'react'
 import _ from 'lodash/fp'
+import { inject } from 'mobx-react'
 import { injectDefaults } from './mobx-react-utils'
 import StripedLoader from './StripedLoader'
 import { autoKey } from './dsl'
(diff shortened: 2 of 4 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 8cc9bbae9a35776233263b404176687be19ef28f
</p>

src/utils/injectTreeNode.js
```diff
@@ -14,40 +16,46 @@ export default (
     style,
   } = {}
 ) =>
-  injectDefaults(({ tree, node, group, path, ...props }) => {
-    node = node || tree.getNode(path)
+  _.flow(
+    inject(({ theme = {} }) => ({
```
stellarhoof: @daedalus28 I though `contextureTheme` a bit redundant. After all, we don't do `contextureTree`

src/utils/injectTreeNode.js
```diff
@@ -14,40 +16,46 @@ export default (
...
+      Loading: (theme.Loading || StripedLoader)(render, style),
```
daedalus28: If we're generalizing this, we probably need a way for custom loaders to have access to `render`, otherwise it will just always show the loading indicator, right?

stellarhoof: I'm not sure I follow. They have access to render. It's being passed as the first argument to the loaders

src/utils/injectTreeNode.js
daedalus28: The idea was that `theme` is more likely to have naming collisions than `tree`, but you're right that it's probably better to be consistent 👍 

At some point, we'll switch to native react context and will reevaluate it anyway

src/utils/injectTreeNode.js
daedalus28: Yup, I was confused because `Loading` and `theme.Loading` are so similar. This does make me think we should reconsider the "loader" API to just be a component instead of a function though

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 1.43.9
```
daedalus28: these versions need updating

PR #235, opened 2019-05-21
Add better validation for TagsInput

Akrion:
Adds check for empty string as well as if the tags is already part of the current set of tags.

### DIFF
--- a/src/layout/TagsInput.js
+++ b/src/layout/TagsInput.js
@@ -7,6 +7,11 @@ import { observer, inject } from 'mobx-react'
 import { Flex } from './Flex'
 import Popover from './Popover'
 
+let isValidTag = (tag, tags) => {
+  let cleanTag = _.toLower(_.trim(tag))
+  return !_.isEmpty(cleanTag) && !_.some(t => _.toLower(t) === cleanTag, tags)
+}
+
 let Tag = observer(({ value, removeTag, tagStyle, RemoveIcon, onClick }) => (
   <span
     className="tags-input-tag"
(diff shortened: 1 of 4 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


Take a look at this commit to see what happened in detail: e966646893117c2bcb385fa1c3429e640ac3d129


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against fe46a9d2fd683a9736b1ccf32b5c1b588bca5462
</p>

src/layout/TagsInput.js
```diff
@@ -7,6 +7,11 @@ import { observer, inject } from 'mobx-react'
 import { Flex } from './Flex'
 import Popover from './Popover'
 
+let isValidTag = (tag, tags) => {
+  let cleanTag = _.toLower(_.trim(tag))
+  return !_.isEmpty(cleanTag) && !_.some(t => _.toLower(t) === cleanTag, tags)
```
daedalus28: We should not assume that all TagsInputs are case insensitive - the simplest thing I can think of is to expose a validateTag function like this as a prop and pass it in on TagsQuery. I suspect we can assume case insensitivity is fine for TagsQuery, but even then that isn't true for exact matches where case _does_ matter. Ultimately, I'm not sure it's even worth it to do this at all because of that.

Akrion: Ok removed the case check.

stellarhoof: A tag being valid is not the only thing `isValidTag` is checking, so the name is a bit off.

A more flexible approach would be to have the option to control this component with an `addTag: (tags, tag) => tags` prop, defaulting to what we've got here.

Also I think trim/lowercase is as good a default as any.

daedalus28: @stellarhoof `addTag` is already a prop since this control doesn't have its own native state

PR #243, opened 2019-06-07
ErrorList improvements

michsa:
1. Ensure no error elements are rendered if ErrorList receives no children. 
   This solves a bug where an empty `<ErrorList block />` was rendering a blank `ErrorBlock`. Also saves us from similar future bugs with `ErrorText`, if we ever change its styling to be more visible.

2. Enable passing props through ErrorList to ErrorText and ErrorBlock. 
   This was already halfway implemented, so I finished the job 🙂


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

<details>
  <summary>ErrorBlock.js</summary>
  <p><a href="https://github.com/smartprocure/contexture-react/blob/feature/error-block-improvements/src/themes/greyVest/ErrorBlock.js" target="_blank">src/themes/greyVest/ErrorBlock.js</a></p>
  
  <strong>'isEmpty' is defined but never used.</strong>
  <code>Line 1: import isEmpty from 'lodash/fp/isEmpty'</code>
</details>
  </td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against d435bd3fe5d31355fa3d3a8bc0512dd75aada0b2
</p>

stories/greyVest/error.stories.js
```diff
@@ -12,6 +12,29 @@ storiesOf('Components (Grey Vest)|Error', module)
...
+  .addWithJSX('Empty', () => (
+    <>
+      <ErrorList />
+      <ErrorList block />
+    </>
+  ))
+  .addWithJSX('Styled', () => (
+    <>
+      <ErrorList style={{ color: 'forestgreen' }}>
+        Ceci n'est pas une error
```
stellarhoof: Uhhh fancy :dancer:

src/themes/greyVest/ErrorList.js
```diff
@@ -1,16 +1,17 @@
...
-  _.map(
-    e =>
-      block ? (
-        <ErrorBlock key={e}>{e}</ErrorBlock>
-      ) : (
-        <ErrorText key={e}>{e}</ErrorText>
-      ),
-    _.castArray(children)
-  )
+let ErrorList = ({ block = false, children, ...props }) =>
```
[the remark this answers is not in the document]
michsa: Okay, but! To use that pattern outside `contexture-react`, we'd have to add `ErrorText` and `ErrorBlock` to our list of exports. I guess that's not really a big deal, but in this particular case reducing the API surface might be worth keeping the flag in there.

On the other hand, no reason not to add an ErrorComponent prop anyway, to allow for custom ones. I just don't want to give up `block` 😣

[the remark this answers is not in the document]
stellarhoof: If you're concerned about exporting `Error{Block,List}` you could use `ErrorList` by default and move `ErrorBlock` to bid-search or do the same with `ErrorList` and export only `ErrorBlock`. This is actually a wider discussion about composition vs configuration that we've had in the past btw. Also don't worry about exporting for now, which is actually another wider discussion around having a solid set of UI primitives that we can nicely compose (because we're not using a component library)

src/themes/greyVest/ErrorList.js
```diff
@@ -3,14 +3,18 @@ import _ from 'lodash/fp'
 import ErrorBlock from './ErrorBlock'
 import ErrorText from './ErrorText'
 
-let ErrorList = ({ block = false, children }) =>
-  _.map(
+let ErrorList = ({ ErrorComponent, block = false, children, ...props }) => {
```
stellarhoof: `block` was introduced recently enough so it's not a big deal if we remove it.

stellarhoof: block was introduced recently enough so it's not a big deal if we remove it.

src/themes/greyVest/ErrorList.js
```diff
@@ -1,20 +1,11 @@
...
-    e =>
-      e ? (
-        <ErrorComponent key={e} {...props}>
-          {e}
-        </ErrorComponent>
-      ) : null,
-    _.castArray(children)
-  )
-}
+let ErrorList = ({ ErrorComponent = ErrorText, children, ...props }) => _.map(
```
stellarhoof: Sorry for the churn:
- Use `React.Children.map` and you won't need `_.castArray`.
- No need to `e ? <ErrorComponent ...> : null`. `e && <ErrorComponent />` works
- I wonder if `Component` is better than `ErrorComponent`. After all, `ErrorList` already hints that we're dealing with errors. What do you think?

michsa: - I changed it back to `_.castArray` when I found out that `React.Children.map` flattens child arrays, such that this will render three ErrorBlocks, rather than two (with the first containing two errors):
  ```
  <ErrorList Component={ErrorBlock}>
    {[['Error 1A', 'Error 1B'], 'Error 2']}
  </ErrorList>
  ```
  Though, that's kind of a silly edge case anyway, and `React.Children.map` is probably safer than `_.castArray`, so I could go either way - might as well use it.

- I didn't think so either, but apparently it is necessary - React complains if you don't explicitly return `null` from a component that doesn't render anything (which is the case for an ErrorList that receives no errors).

- Good point, changed 🙂

src/themes/greyVest/ErrorList.js
```diff
@@ -1,16 +1,12 @@
...
-let ErrorList = ({ block = false, children }) =>
-  _.map(
-    e =>
-      block ? (
-        <ErrorBlock key={e}>{e}</ErrorBlock>
-      ) : (
-        <ErrorText key={e}>{e}</ErrorText>
-      ),
-    _.castArray(children)
+let ErrorList = ({ Component = ErrorText, children, ...props }) =>
```
daedalus28: At this point, this isn't an `ErrorList`,  it's basically just a `List` that happens to default to `ErrorText`. Something this flexible starts to become of questionable value. The benefit of this:

```jsx
<List Component={SomeComponent} someProp="value">
  {['wrap', 'these']}
</List>
```
over this:

```jsx
_.map(
  x => <SomeComponent someProp="value">{x}<SomeComponent>
  ['wrap', 'these']
)
```

isn't super clear, except that it's a thing you need to know now and gets pretty close to a "foreach" component (which is generally discouraged by the react community as it starts to get into a template language kinda thing instead of just being plain js). I suppose the ideal case for it might be wrapping actual JSX components:

```jsx
<List Component={SomeComponent} putThisProp="onAllChildren">
  <Wrap />
  <These />
  <Components />
</List>
```

which I suppose is better than

```js
_.map(
  x => <SomeComponent putThisProp="onAllChildren">{x}<SomeComponent>
  [
    <Wrap />,
    <These />,
    <Components />
  ]
)

```

Still, it definitely seems like a pure preference thing that doesn't have a clear answer. This seems like something we should open up for discussion by the rest of the team 🤔. I don't feel strongly about it, but it does strike as me something we might not want to do. If we do proceed, we should definitely factor out List and make ErrorList just be `defaultProps` on List.

stories/greyVest/error.stories.js
```diff
@@ -2,20 +2,43 @@ import React from 'react'
...
 import { Flex } from './../../src/layout/Flex'
-import { Box, ErrorList, Input } from './../../src/themes/greyVest'
+import { Box, ErrorList, ErrorBlock, Input } from './../../src/themes/greyVest'
 
 storiesOf('Components (Grey Vest)|Error', module)
   .addDecorator(decorator)
   .addWithJSX('Text', () => <ErrorList>I am an error</ErrorList>)
   .addWithJSX('Block', () => (
-    <ErrorList block>
+    <ErrorList Component={ErrorBlock}>
```
daedalus28: This does seem like a less friendly API. I think at the end of the day, these are really just two different components, ErrorList and ErrorBlocks.

src/themes/greyVest/ErrorList.js
michsa: I was about to say that we don't have anything that would justify a generic `List` component, though actually, we kinda do -- there's currently no list markup anywhere in contexture-react, but one could argue that there *should* be, for semantic reasons, and a `List` that wraps its children in `<li>` would accomplish that pretty well.

(I doubt anyone will argue that it's worth our time to pursue semantic HTML standards right this second, but since it might one day be necessary for accessibility stuff, might as well bring it up where it's relevant.)

In defense of `ErrorList`, it seems to me that the intent, based on the majority of its actual usage (eg. autoform fields), was less "list of errors" and more "an error (but possibly more than one)". In practice, it almost always holds only one error -- could be something to keep in mind if we do decide to use `List`.

PR #247, opened 2019-06-13
Rename `tree` props in QueryBuilder

michsa:
Renames all `tree` props (which are actually nodes) in QueryBuilder to `node`. 

This is a prerequisite to mopping up the rest of ContextureClientBridge, which creates a wrapper around the contexture tree and passes it to these components as `root`. After that's gone, we'll be passing a new `tree` prop with a reference to the contexture tree instead.

### DIFF
--- a/src/queryBuilder/Group.js
+++ b/src/queryBuilder/Group.js
@@ -15,7 +15,7 @@ let { background } = styles
 let GroupItem = FilterDragSource(args => {
   let {
     child,
-    tree,
+    tree: node,
     index,
     state,
     root,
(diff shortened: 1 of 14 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against eaa120233178698b643a011ba19f1e6de061a7f3
</p>

src/queryBuilder/Group.js
```diff
@@ -15,32 +15,32 @@ let { background } = styles
...
-        ...(index === tree.children.length - 1 &&
+        ...(index === node.children.length - 1 &&
           !root.adding && { background }),
       }}
     >
-      {!(isRoot && tree.children.length === 1) && (
+      {!(isRoot && node.children.length === 1) && (
         <Operator
-          {...{ tree, child, root, parentTree, index, parent: state }}
+          {...{ node, child, root, parentNode, index, parent: state }}
```
daedalus28: Given that we have a lot of instances of `node` and `child`, maybe we should go with `parent` instead of `parentNode`? Seems like we should either do that or rename `child` to `childNode`, but I think `parent` and `child` without adding node on the end are probably clear enough names

PR #250, opened 2019-06-17
Remove ContextureClientBridge

michsa:
- Replace all `root` props in query builder components with `tree` props containing a reference to the contexture tree
- Replace all calls to `root` methods in query builder components with the corresponding contexture-client actions
- Move/add utility methods to `src/utils/tree.js` -- everything ContextureClientBridge did that was more than simple syrup over a contexture tree action is now here. It's not ideal and I think most of it can/should still be refactored (particularly `indent`), but this PR is already pretty big, so I decided to leave it for now.

One benefit of all this is the ability to add functionality to FilterList (basic search) that was previously restricted to QueryBuilder, because it relied on methods in ContextureClientBridge instead of in contexture-client itself. Changing a node's type (via [the new `replace` action](https://github.com/smartprocure/contexture-client/pull/99)) is one example, but anything that can be done in QueryBuilder is now possible to implement in FilterList too. 🙂

PR #253, opened 2019-06-24
Add type options to title popover in FilterList

michsa:
Closes an issue 🎉

Branched from #250, and depends on it for `newNodeFromType` (I think that's actually the only thing, though...)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


Take a look at this commit to see what happened in detail: 3f40f0eca3b23a0dcaabb86c2cbb0f960e01fb88


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 0e5e18221e85f05e789abea26fb09c63e4538a97
</p>

src/FilterList.js
```diff
@@ -39,11 +40,43 @@ export let Label = inject(_.pick('tree'))(
...
+                  ),
+                  F.autoLabelOptions(
+                    _.without(
+                      [node.type],
+                      _.get([node.field, 'typeOptions'], fields)
+                    ) || []
+                  )
+                )}
+                <div
+                  style={{ borderBottom: '1px solid #eee', margin: '4px -5px' }}
```
daedalus28: This probably shouldn't be in the core components and should be a themeable thing, e.g. something like `className="dropdown-option-separator"` or whatever and then themed in grey vest

stellarhoof:
@michsa Can you add a screenshot?

michsa:
@stellarhoof here is it in grey vest:

Screen Shot 2019-06-26 at 11 32 56 AM

And in blueberry (unstyled):

Screen Shot 2019-06-26 at 11 33 39 AM

stellarhoof:
I'm not sure what the policy on new files is, but my first instinct would be to keep `FilterOptions` inside `FilterList`, unless you've got plans for it somewhere else.

Also what do you think about `FilterActions`?

Otherwise looks good.

Oh PS: Why the dash at the beginning of the filter types?

michsa:
Yeah, `filterActions` might be better. And you're probably right about just keeping it inside `FilterList` - we can always factor it out later if necessary. Plus, this is actually only half the feature - there will be also an option to change the node's field, which should open a modal picker. Making that work from inside a popover is pretty problematic, but if it's all inside `FilterList` hopefully it will be easier.

And the dash wasn't my idea (but I do think it looks nice)

src/FilterOptions.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
+import _ from 'lodash/fp'
+import F from 'futil-js'
+import { newNodeFromType } from './utils/search'
+
+let FilterOptions = ({ node, tree, fields, Item }) => (
```
daedalus28: agreed with @stellarhoof - this doesn't mean much outside of `FilterList`, and `FilterActions` is probably a better name

src/FilterOptions.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react'
```
stellarhoof: You forgot to remove this file :D

michsa: 🤦‍♀

src/themes/greyVest/index.js
```diff
@@ -490,6 +490,23 @@ export let GVStyle = () => (
         padding-top: 30px;
       }
 
+      .filter-options-popover {
```
stellarhoof: It'd be nice if these were `filter-actions-*` as well

PR #255, opened 2019-06-26
Add *field* options to title popover in FilterList

michsa:
For this case, I needed a ModalPicker that would respond to a stateLens hooked up to a button elsewhere, rather than managing the state and button internally - since, as we discussed, the button to toggle the modal is inside a popover, and the modal should be outside.

So, I factored out the modal/picker part of `ModalPicker` as `ExternalModalPicker`. I personally like this solution, but would appreciate some feedback, particularly from @stellarhoof (and especially on the name, which... probably isn't the best) 😅


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


Take a look at this commit to see what happened in detail: b0e1bd3bb5a6a5fb984fb5b99d63d869427b0812


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against e8343249f96b33d3acd713e6b06d9c7875bc904e
</p>

src/layout/index.js
```diff
@@ -5,7 +5,7 @@ import Modal from './Modal'
 import Popover from './Popover'
 import SpacedList from './SpacedList'
 import TextHighlight from './TextHighlight'
-import { ModalPicker } from './Pickers'
+import { ModalPicker, ExternalModalPicker } from './Pickers'
```
stellarhoof: Bogus import?

src/FilterList.js
```diff
@@ -2,14 +2,27 @@ import React from 'react'
...
+import { Flex, Dynamic, Popover, Modal, NestedPicker } from './layout'
+import { fieldsToOptions } from './FilterAdder'
 import { withStateLens } from './utils/mobx-react-utils'
 import InjectTreeNode from './utils/injectTreeNode'
 import DefaultIcon from './DefaultIcon'
 import { bdJoin } from './styles/generic'
 import { newNodeFromType } from './utils/search'
 
-export let FilterActions = ({ node, tree, fields, Item }) => (
+let FieldPicker = ({ options, isOpen, onChange }) => (
```
stellarhoof: You can save yourself moving state around (`isOpen` in this case) if you move the button here and use `withState` here as well.

daedalus28: The button needs to be separate because the modal has to be detached from the button (so that it's not rendered inside the popover)

stellarhoof: Don't we use a portal for the modal? Why does it render inside the popover?

michsa: It's not rendered inside the popover, but it is killed when the popover closes, which (if we did put the modal inside) would happen as soon as you try to interact with the modal, or even as soon as you click the item to open the modal in the first place.

It's possible (in theory) to hack around the normal popover functionality to avoid these problems, but it's not worth it - best to just keep the modal outside.

src/layout/index.js
michsa: Sorry, forgot to take that out - this is what I had before:

https://github.com/smartprocure/contexture-react/pull/255/files/b0e1bd3bb5a6a5fb984fb5b99d63d869427b0812

My idea was to just export a `ModalPicker` component that doesn't include its own button and state, so it can be managed... `External`ly 🙃

Then Sam pointed out that this was probably unnecessary overhead so I threw in a commit to show the difference if it were inlined in `FilterList` instead. And yeah, I guess it is kind of unnecessary... I still kinda like it, but neither of us feels strongly about it. What do you think?

src/FilterList.js
stellarhoof: Yeah even with portals if the parent unmounts so does the child (Modal in this case).

> It's possible (in theory) to hack around the normal popover functionality

I wouldn't call our current popover "normal" lol. The name on this https://github.com/airbnb/react-outside-click-handler is misleading. It should just handle outside clicks but it closes the popover when it's clicked inside :(

stellarhoof: I would add a comment somewhere though, because someone could have the same question about why isn't the state closer to where it's used.

src/layout/index.js
stellarhoof: I like less components because I have to come up with less names. Also Modal + Anything is not really worth exporting in a generic library like this imo since that composition is trivial to do.

stellarhoof:
Looks good

michsa:
Just realized this has been bugged this whole time -- something isn't watching state correctly, so changing a filter's type after the field (or vice versa) fails to update either the label text or the dynamic child component 

So, yeah, please don't merge yet 😓

michsa:
So adding `observer` to FilterActions fixed the update issues, but there are still some odd cases with changing fields and types. Specifically, it seems that if there are extra properties on a node (relevant to the type it was initialized with, like `min` and `max` for numbers), and you change its type to one that doesn't have them, it can cause errors that prevent the dynamic component for the new type from loading, or otherwise break the node.

Turns out that this also happens in QueryBuilder, so it's not actually an issue with this PR (though I guess now it'll be more visible).

PR #258, opened 2019-07-03
QueryWizard components 🧙‍♂️

michsa:
Preview:

Screen Shot 2019-07-03 at 1 25 56 PM

Screen Shot 2019-07-03 at 1 26 18 PM

Still technically WIP (~~the modals need some styling~~, and I'm sure it will see a few tweaks before it's done), but it seems to work and looks somewhat okay, so I figured I'd throw up the PR 🙂

Modals (with extremely placeholder "friendly" text):

Screen Shot 2019-07-05 at 8 00 49 AM


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


Take a look at this commit to see what happened in detail: c60450d25703b9ee1ab61a268cafb4ad3f66cfe5


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 878eee19896b0bbb6e0c4e994135c7e0c90077e7
</p>

src/queryWizard/WizardGroup.js
```diff
@@ -0,0 +1,119 @@
...
+))
+
+let FieldDescription = observer(({ node }) => (
+  <div>
+    {node.fieldDescription} {node.typeDescription}
+  </div>
+))
+let WizardItem = ({
+  node,
+  tree,
```
daedalus28: we should follow the convention you established on your QueryBuilder PR to call this root and not tree

src/queryWizard/WizardGroup.js
```diff
@@ -0,0 +1,119 @@
...
+}) => (
+  <Box {...{ nested, className }} nodeJoinColor={styles.joinColor(node)}>
+    {_.map(
+      child =>
+        child.children ? (
+          <WizardGroup
+            key={child.path}
+            nested
+            {...{
+              tree,
```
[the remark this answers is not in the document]
michsa: I'm personally not a fan of that pattern - it makes it much less obvious what actually needs to be passed. I know you can still check the props in the component's definition, but when first reading (or re-reading), I find it really helps to see it right inside the parent.

This is especially true in situations where a node forwards props its children without destructuring them for its own use. Eg, you can't tell just from looking at `A` that it's expected to take a `Bar` prop and pass it along to `B`:

```jsx
let A = props => {
  let { foo } = props
  return (<B {...props} />)
}

let B = ({ foo, Bar}) => (<Bar>{foo}</Bar>)
```

This happens sometimes with our component-passing pattern (especially in QueryBuilder), and it definitely makes it harder to reason about when you're still familiarizing yourself with the code

src/queryWizard/QueryWizard.js
```diff
@@ -12,6 +12,7 @@ export default withStateLens({ current: 0 })(
   observer(
     ({
       tree,
+      path,
```
michsa: @daedalus28 regarding the `path` prop and roundtripping with other search formats -- my thinking, to still allow arbitrary numbers of `WizardStep`s (and avoid having to match some hardcoded collection of `stepN` keys to render step groups), is to have QueryWizard handle separate paths for:
- "steps", a group node whose children are step groups, where the first is titled **"Search for ${schemaLabel} by..."** in QueryWizard and the rest are titled **"And..."** (which we can show together in a non-editable box in the other modes), and
- "filters", the path to the editable group in the other modes (so, typically `root/criteria/filters`), which is rendered as the final **"Narrow Your Results"** step in QueryWizard

I can think of a few way to handle this, on the spectrum of simple/clear vs easy/flexible:

1. `stepsPath` and `filtersPath` props
2. `path` prop for the steps path, `filtersPath` (or `additionalFilters` or something) for the filters path
4. All three of `stepsPath`, `filtersPath`, and `path`; if the latter is used, it looks for `steps` and `filters` children on the node at that path

What do you think?

src/queryWizard/QueryWizard.js
```diff
@@ -0,0 +1,54 @@
...
+import F from 'futil-js'
+import React from 'react'
+import { observer } from 'mobx-react'
+import DefaultCheckButton from '../layout/CheckButton'
+import DefaultModal from '../layout/Modal'
+import DefaultIcon from '../DefaultIcon'
+import WizardStep from './WizardStep'
+import { withStateLens } from '../utils/mobx-react-utils'
+
+export default withStateLens({ current: 0 })(
```
stellarhoof: Make sure to set the displayName of default exports like these. Otherwise the name won't show up on React devtools.

stellarhoof: Just use react state instead of `withStateLens`. You won't need observable either since it doesn't seem like you're observing anything worth a re-render.

src/queryWizard/WizardStep.js
```diff
@@ -0,0 +1,91 @@
...
+import { observer } from 'mobx-react'
+import DefaultCheckButton from '../layout/CheckButton'
+import DefaultModal from '../layout/Modal'
+import { Flex } from '../layout/Flex'
+import WizardGroup from './WizardGroup'
+
+// Observes node, so we can activate the Continue button if it (or any child) has a value.
+// We don't observe on WizardStep because then it would rerender its children when `node`
+// changes, which unfocuses query inputs as soon as the first character is entered.
+let Buttons = observer(
```
stellarhoof: I'd move this to `QueryWizard`. Less props for `WizardStep` which has lots.

src/FilterButtonList.js
```diff
@@ -0,0 +1,139 @@
...
+  Flex,
+  CheckButton as DefaultCheckButton,
+  Modal as DefaultModal,
+  Popover as DefaultPopover,
+} from './layout'
+import DefaultIcon from './DefaultIcon'
+import InjectTreeNode from './utils/injectTreeNode'
+import styles from './styles'
+
+export let DefaultMissingTypeComponent = InjectTreeNode(({ node = {} }) => (
```
daedalus28: This is duplicated from other parts of the code base, so let's factor it so it's not copied and pasted

src/FilterButtonList.js
```diff
@@ -0,0 +1,139 @@
...
+  <div>
+    Type <b>{node.type}</b> is not supported (for key <i>{node.key}</i>)
+  </div>
+))
+
+let FieldDescription = observer(({ node }) => (
+  <div className="filter-description">
+    {node.fieldDescription} {node.typeDescription}
```
daedalus28: We should avoid putting these on `node` because they're display only and don't need to be sent to the server or persisted in the DB. you can either drive these as props to FilterButtonItem (which can be set via something like `mapNodeToProps`) or look up description off of the `field` from the schema an override `description` by simply changing the type components that are passed in at the end user level

src/FilterButtonList.js
```diff
@@ -0,0 +1,139 @@
...
+          <Button primary onClick={F.off(modal)}>
+            Done
+          </Button>
+        </div>
+      </Modal>
+    </>
+  )
+}
+
+let Box = ({ nodeJoinColor, children, nested, className }) => (
```
daedalus28: This name is a little ambiguous since there is a GreyVest `Box` component. Maybe something like `GroupBox` or `TreeBox`?

src/FilterButtonList.js
```diff
@@ -0,0 +1,139 @@
...
+                Button,
+                CheckButton,
+                Icon,
+                Modal,
+                Popover,
+              }}
+              label={mapNodeToLabel(child, fields)}
+            />
+          ),
+        _.getOr([], 'children', node)
```
daedalus28: `_.map` handles null/undefined, so you should be able to just do `node.children` since I don't think `node` itself can be undefined

src/FilterButtonList.js
```diff
@@ -0,0 +1,139 @@
...
+    alignItems="center"
+    style={{ borderColor: nodeJoinColor }}
+  >
+    {children}
+  </Flex>
+)
+
+let FilterButtonList = observer(
+  ({
+    node,
```
daedalus28: if this is not using `injectTreeNode`, are we expecting this to be explicitly passed from the client? I'm ok with it in theory, but we do have precedent of supporting `path` + `tree` instead of requiring end users to do `node={tree.getNode(path)}`

src/queryWizard/AccordionWizard.js
```diff
@@ -0,0 +1,39 @@
+import _ from 'lodash/fp'
+import F from 'futil-js'
+import React from 'react'
+import DefaultIcon from '../DefaultIcon'
+import WizardStep from './WizardStep'
+
+let splitKeys = _.curry((keys, obj) => [_.pick(keys, obj), _.omit(keys, obj)])
```
daedalus28: I think you might be looking for [`_.partition`](https://lodash.com/docs/4.17.11#partition)

michsa: I was - not sure how I missed it 🤦‍♀

src/FilterButtonList.js
michsa: FilterList does this also, so I figured there must've been some precedent for it as a failsafe. It's technically possible for a user to omit `path` and pass in an undefined node (and expect it to just render nothing, rather than throwing) -- but in that case, `_.get('children', node)` should work just as well.

I guess it's really a question of whether filter components should support undefined nodes, or not (in which case, the "can't read `children` of undefined" exception is reasonably informative imo)

src/queryWizard/AccordionWizard.js
michsa: actually, `_.partition` won't work - it only passes `value` to the predicate, so we still need a new function for splitting on keys.

(I also played around with `reduce` to avoid mapping twice, but it's significantly less readable for what's likely a completely unnoticeable performance gain, so I guess I'll just stick with `_.pick` and `_.omit`)

src/FilterButtonList.js
michsa: These were supposed to be placeholders for the functionality you're describing - it completely slipped my mind that I hadn't implemented it yet 🤦‍♀

For the wizard stuff, I think we'll need all of the above for maximum granularity. So `label`/`description` props from `mapNodeToProps` take precedence (for specific overrides), then the value of `mapNodeToLabel`/`mapNodeToDescription`, then various fallbacks like `node.field` (for the label). 

This is particularly relevant for the description stuff; we'll need to be able to knit looked-up field and type descriptions together in a friendly-enough way, and `mapNodeToDescription` allows us to do that while still having the option to set descriptions on a per-node basis with `mapNodeToProps`.

src/FilterButtonList.js
```diff
@@ -0,0 +1,152 @@
...
+  Button,
+  CheckButton,
+  MissingTypeComponent,
+  Modal,
+}) => {
+  let mappedProps = mapNodeToProps(node, fields)
+  let modal = F.stateLens(React.useState(false))
+  let title = // we really need a title, so here's every possible fallback
+    _.get('label', mappedProps) ||
+    mapNodeToLabel(node, fields) ||
```
daedalus28: i don't think we need to explicitly support mapNodeToLabel and mapNodeToDescription, just mapNodeToProps is probably fine

stories/queryWizard/utils.js
```diff
@@ -0,0 +1,26 @@
...
+
+let differentFirst = (firstCase, normalCase) => (acc, i, list) =>
+  i === 0
+    ? _.iteratee(firstCase)(acc, i, list)
+    : _.iteratee(normalCase)(acc, i, list)
+
+let hasPunctuation = F.testRegex(/.*(?=[.!?]$)/)
+
+let punctuate = (punctuation = '.') =>
+  F.when(F.overNone([_.isEmpty, hasPunctuation]), F.append(punctuation))
```
daedalus28: I'm not sure I understand why we need all of this. Why not just have periods in the description strings? 🤔

src/FilterButtonList.js
daedalus28: also mergeOverAll can help compose mapNodeToProps functions

stories/queryWizard/utils.js
michsa: Uh... "just in case" 😅

I wanted to make double dog sure that stitched-together field and type descriptions would look nice and accessible in as many situations as possible -- but, yeah, I guess improperly-formatted description text is trivially preventable as far as edge cases go, so this may have been a little overboard

src/layout/StepsAccordion.js
```diff
@@ -0,0 +1,112 @@
...
+  totalSteps,
+  currentStep,
+  stepTitle = '',
+  isRequired = false,
+  onSubmit,
+  children,
+}) => {
+  let isOpen = F.view(currentStep) === step
+  return (
+    <div className={`wizard-step ${className ? className : ''}`} style={style}>
```
daedalus28: This class name should probably be accordion step and not wizard step

stories/queryWizard/config.js
```diff
@@ -0,0 +1,111 @@
...
+      label: 'Friendly Group',
+      type: 'group',
+      join: 'and',
+      children: [
+        {
+          type: 'tagsQuery',
+          key: 'friendly node',
+          label: 'Friendly Node',
+          typeDescription: 'Enter some tags',
+          fieldDescription: 'Search for stuff',
```
daedalus28: These should never be on contexture nodes

stories/queryWizard/examples.stories.js
```diff
@@ -0,0 +1,81 @@
...
+})
+
+let story = QueryWizard => () => (
+  <QueryWizard
+    tree={tree}
+    path={['root']}
+    fields={fields}
+    mapNodeToProps={componentForType(TypeMap)}
+    mapNodeToLabel={(node, fields) => _.get([node.field, 'label'], fields)}
+    mapNodeToDescription={mapNodeToDescription(types)}
```
daedalus28: These shouldn’t still be here

stellarhoof:
Looks good. I can only comment on the code though, not on whether the PR as a whole is ok.

stories/layout.stories.js
```diff
@@ -119,3 +117,18 @@ storiesOf('Components (Unthemed)|Layout', module)
       <CheckButton>This one is off</CheckButton>
     </>
   ))
+  .addWithJSX('StepsAccordion', () => (
+    <StepsAccordion>
+      <div isRequired={true}>
```
daedalus28: It's a little weird to pass props to a component that it wouldn't support and then remove them internally. It seems like it would be cleaner to have an `AccordionStep` component instead.

src/themes/greyVest/index.js
```diff
@@ -861,3 +942,27 @@ export let QueryBuilder = defaultProps({ Button, MissingTypeComponent })(
...
+  CheckButton,
+  Icon,
+  Modal,
+  className: 'gv-filter-button-list',
+})(BaseFilterButtonList)
+
+export let StepsAccordion = defaultProps({
+  Button,
+  Icon,
+  className: 'gv-query-wizard',
```
daedalus28: is this right? I see you are also adding it to QueryWizard

stories/queryWizard/config.js
daedalus28: so let's take them out :)

stories/queryWizard/examples.stories.js
daedalus28: so let's remove them :)

src/layout/StepsAccordion.js
```diff
@@ -0,0 +1,112 @@
...
+      <Flex alignItems="center" justifyContent="space-between">
+        <Flex alignItems="center">
+          <h1>
+            <span className="step-number">Step {step + 1}</span>
```
daedalus28: Just a thought - it occurs to me that this might not always be desirable, so maybe it shouldn't be the responsibility of this component to add `Step n`. I don't feel super strongly about this though so I'm open - just wanted to bring it up

stories/queryWizard/examples.stories.js
michsa: sorry - I could swear I removed these locally at one point, and they somehow snuck back in

it is a mystery 👻

stories/queryWizard/config.js
michsa: this one was pure oversight :(

src/themes/greyVest/index.js
michsa: ah, this was a holdover from splitting out StepsAccordion. I changed it to `gv-steps-accordion` and removed the QueryWizard one, since QueryWizard just returns a StepsAccordion anyway.

I suppose we could add a wrapper div to QueryWizard so we can give it separate class/style props, but I don't think there's any real use case for that now (and it would be trivial to add if it ever becomes necessary)

PR #260, opened 2019-07-08
Add `childSpacing` prop to Flex

michsa:
Expands the `Flex` component API to include a `childSpacing` prop. When set, this intersperses spacer elements between `Flex`'s children:
```jsx
<div style={{ flex: `0 1 ${childSpacing}` }} />
```

This creates a div that is by default (and at most) the width of `childSpacing`, but is able to shrink if the other elements require more space.

⚠️ I know this somewhat violates our current convention for Flex (including only props that directly translate to flexbox styles), **but**:
1. in practice, we usually do want to have some space between flex children
2. intuitively, "space between items" should be a property of the presentational element containing the items, rather than a property of the individual items that can't account for being first or last in the list (which is why I think it makes sense to include it here, in spite of the convention)
3. as far as I'm aware, this is literally the least-bad way to add spacing between dynamically generated flex children

(An alternate not-that-bad strategy would be to use a CSS-in-JS framework that supports pseudoselectors and set:
```es
'> *': {
  margin: childSpacing,
  '&:first-child': {
    margin-left: `-${childSpacing}`,
  },
  '&:last-child': {
    margin-right: `-${childSpacing}`,
  },
}
```

...but this way is simpler, works just as well (if not better), and is already doable with our current tools.)

### DIFF
--- a/src/layout/Flex.js
+++ b/src/layout/Flex.js
@@ -1,4 +1,8 @@
 import React from 'react'
+import F from 'futil-js'
+import _ from 'lodash/fp'
+
+let defaultPx = prop => (_.isNumber(prop) ? `${prop}px` : prop)
 
 export let Flex = ({
   as: Component = 'div',
@@ -6,8 +10,10 @@ export let Flex = ({
   alignItems,
   alignContent,
   justifyContent,
+  childSpacing,
   wrap = false,
   column = false,
+  children,
   ...props
 }) => (
   <Component
@@ -21,5 +27,12 @@ export let Flex = ({
       ...style,
     }}
     {...props}
-  />
+  >
+    {childSpacing
+      ? F.intersperse(
+          () => <div style={{ flex: `0 1 ${defaultPx(childSpacing)}` }} />,
+          children
+        )
+      : children}
+  </Component>
 )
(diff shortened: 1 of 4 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 640237147c7e7135900fd618ed8e0dc1f4bf1941
</p>

src/layout/Flex.js
```diff
@@ -21,5 +27,12 @@ export let Flex = ({
       ...style,
     }}
     {...props}
-  />
+  >
+    {childSpacing
+      ? F.intersperse(
+          () => <div style={{ flex: `0 1 ${defaultPx(childSpacing)}` }} />,
```
stellarhoof: You don't need a function. `intersperse` also accepts a value

PR #261, opened 2019-07-12
Adding one line support for the TagsInput

Akrion:
As per https://github.com/smartprocure/bid-search/issues/2801 this makes the TagsInput always be one line unless clicked on. Then it retains its full height instead of being limited via `max-height`.

Also changed the `padding` to `2` since it hides better the color lines of the tags.

End result:

image

When clicked:

image

Also focused the input on the container click so that user `does not` have to click 2 times (as it is the case currently) to enter tags.

### DIFF
--- a/src/layout/TagsInput.js
+++ b/src/layout/TagsInput.js
@@ -109,10 +109,10 @@ let TagsInput = withState('state', 'setState', () =>
             }}
           >
             <Flex
+              wrap
+              alignItems="center"
               style={{
                 cursor: 'text',
-                alignItems: 'center',
-                flexWrap: 'wrap',
                 height: '100%',
                 padding: 2,
               }}
--- a/src/layout/TagsInput.js
+++ b/src/layout/TagsInput.js
@@ -6,6 +6,7 @@ import { observable } from 'mobx'
 import { observer, inject } from 'mobx-react'
 import { Flex } from './Flex'
 import Popover from './Popover'
+import OutsideClickHandler from 'react-outside-click-handler';
 
 let isValidTag = (tag, tags) => {
   let cleanTag = _.trim(tag)
@@ -63,6 +64,7 @@ let TagsInput = withState('state', 'setState', () =>
     currentInput: '',
     selectedTag: null,
     popoverOpen: false,
+    isOneLine: true,
   })
 )(
   observer(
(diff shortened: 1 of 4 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


Take a look at this commit to see what happened in detail: 35a2afa6944578d18b91a6920fbb10056b787590


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 74652361999c026d7a7d48423048294c3ae144db
</p>

Akrion:
@stellarhoof 

1. No it should be for all since as per the ticket mentioned in Spark (https://github.com/smartprocure/bid-search/issues/2801) we have the exact same issue with the `Not` field.

2. This is an issue not related to this PR. I actually fixed that in Spark so there this does not happen. This is related to how the search bar is setup in the demo here. See in beta/prod:

image

3. Just click on the white space of the search box or on any of the tags. Clicking on the tags would also focus on the search box etc

stellarhoof:
Might as well fix 2. here. It's pretty easy, just add `style={{ alignSelf: 'start' }}` to the div with the `gv-button-group` class

For 3. I know I can click on the whitespace but there's very little and it's not apparent that clicking there would do anything. In the main-app gif there's enough space to the left so that is clear that you can click there.

Akrion:
The little space you are talking about in 3 is simply due to the edge case where you added words that ended up being perfectly aligning in terms of width to each other. As you can see from both my screenshots that would rarely happen. Not sure if this is worth any time invested at this point. If it becomes a need then it should be specified in better requirements. Current ones are "Make it as mainapp" .

michsa:
The whitespace alone isn't that bad of an issue, but it conspires with some other factors to make it extremely non-obvious that there are additional tags under the cut:

tags

Specifically,
1. We can't rely on the appearance of a scrollbar to convey that there's overflow content, because scrollbars are hidden by default on MacOS
2. Partially-visible content, like the top few pixels of the next line of tags, *does* communicate pretty well that there's overflow (even if it's sloppy), but the added padding takes away that benefit

Also, the search box looks the same as the non-interactible Box containers, so it needs a hint that the contents are interactible unless we do something else to distinguish it visually (like adding a border around the input, which is what the old app did). Before, this was accomplished by the `Search...` placeholder text -- but if that's hidden along with the overflowed tags, the *only* affordance that communicates interactibility is the input cursor when you hover over the whitespace, and that's really not enough.

Akrion:
@michsa that is fair for the macOS. Updated with styling of the scrollbar (which we did have in mainapp and I forgot to move here):

image

src/layout/TagsInput.js
```diff
@@ -79,87 +81,109 @@ let TagsInput = withState('state', 'setState', () =>
...
-                margin: 3,
-                minWidth: 120,
-              }}
-              onChange={e => {
-                state.currentInput = e.target.value
+                cursor: 'text',
+                alignItems: 'center',
+                flexWrap: 'wrap',
+                height: '100%',
+                padding: 2,
```
michsa: FYI, the `alignItems` and `flexWrap` properties are now part of our Flex component API, so you could also do:

```jsx
<Flex 
  wrap
  alignItems="center"
  style={{
    cursor: 'text',
    height: '100%',
    padding: 2 
  }}
>
```

Akrion: Sure it seems to work so why not. Changes made.

PR #266, opened 2019-07-18
Use tree.replace instead of tree.mutate to change a node's field

stellarhoof:
With this we'll change the exampleType component correctly as well as take into account custom default node props

### DIFF
--- a/src/FilterAdder.js
+++ b/src/FilterAdder.js
@@ -2,21 +2,13 @@ import _ from 'lodash/fp'
 import React from 'react'
 import { observer } from 'mobx-react'
 import InjectTreeNode from './utils/injectTreeNode'
-import { DefaultNodeProps } from './utils/schema'
+import { newNodeFromField } from './utils/search'
 
 export let fieldsToOptions = _.map(x => ({ value: x.field, ...x }))
 
 let getGroupFields = node => _.map('field', _.getOr([], 'children', node))
 
-let FilterAdder = ({
-  tree,
-  node,
-  path,
-  fields,
-  Picker,
-  uniqueFields,
-  defaultNodeProps = DefaultNodeProps,
-}) => {
+let FilterAdder = ({ tree, node, path, fields, Picker, uniqueFields }) => {
   let options = fieldsToOptions(fields)
   if (uniqueFields) {
     options = _.reject(x => _.includes(x.field, getGroupFields(node)), options)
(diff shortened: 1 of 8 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 57cc581ff9f4084e3312ce079400246257354425
</p>

src/FilterAdder.js
```diff
@@ -2,21 +2,13 @@ import _ from 'lodash/fp'
...
 let getGroupFields = node => _.map('field', _.getOr([], 'children', node))
 
-let FilterAdder = ({
-  tree,
-  node,
-  path,
-  fields,
-  Picker,
-  uniqueFields,
-  defaultNodeProps = DefaultNodeProps,
```
stellarhoof: @daedalus28 Do we want this? Nothing uses it.

daedalus28: also even if it does not, contexture-react shouldn't dictate that's the only way to add default node props

src/utils/search.js
```diff
@@ -18,13 +18,15 @@ export let newNodeFromType = _.curry((type, fields, node) => ({
...
 }))
 
-// In the future, we will have logic here to handle intelligently updating
-// a node's types when its field is mutated. When that happens, we'll also
-// need to pass the `fields` schema to look up typeOptions for the current
-// and future field. Could also be a node transformer like newNodeFromType
-// if we decide to use Tree.replace() for fields instead of Tree.mutate().
-export let changeNodeField = (Tree, node, field) => {
-  Tree.mutate(node.path, { field })
+export let newNodeFromField = ({ key, field, fields, ...rest }) => {
```
michsa: personally, I don't prefer this style of argument passing unless it's really necessary - I'd probably just have done `(field, fields, rest = {})`, with `key` in the rest object (or not at all, see the next comment)

it's really just a matter of taste, though, so it's up to you 🙂

src/FilterAdder.js
stellarhoof: > contexture-react shouldn't dictate that's the only way to add default node props

Adding it back to merge this PR, but this flexibility is only supported here. We just go with a vanilla `defaultNodeProps` everywhere else

stellarhoof: > contexture-react shouldn't dictate that's the only way to add default node props

Why here and not in all the other places we're using it? This is the only place where we offer overriding it and we're not using it anyway.

daedalus28: ok, we can always add this back later if need be then

PR #269, opened 2019-07-22
Wizard fixes

michsa:
Wizard fixes on the contexture-react side 🧙‍♂️

~~Most importantly, this adds a `useLoader` flag to the config parameter for `injectTreeNode`, and the ability to turn off the loader wrapper by setting it to `false` - which stops those components from rerendering when their node becomes marked for update.~~

Per feedback, everything to do with `injectTreeNode` has been removed from here and included in a separate PR. There are now only two changes:

- StepsAccordion: make clicking on a step's title fold/unfold that step
- QueryWizard: add missing `key` prop so React stops complaining

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 1.54.6
+* Tweaks and fixes to QueryWizard & StepsAccordion
+
 # 1.54.5
 * Add support for type labels to type selection options in FilterList/QueryBuilder
 * ResultTable: Pass additional props to Row
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "contexture-react",
-  "version": "1.54.5",
+  "version": "1.54.6",
   "description": "React components for building contexture interfaces",
   "main": "dist/index.js",
   "scripts": {
--- a/src/layout/StepsAccordion.js
+++ b/src/layout/StepsAccordion.js
@@ -53,18 +53,19 @@ export let AccordionStep = ({
       className={`accordion-step ${className ? className : ''}`}
       style={style}
     >
-      <Flex alignItems="center" justifyContent="space-between">
+      <Flex
+        alignItems="center"
+        justifyContent="space-between"
+        onClick={F.sets(isOpen ? -1 : step, currentStep)}
+        style={{ cursor: 'pointer' }}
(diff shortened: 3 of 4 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


Take a look at this commit to see what happened in detail: 83f9372c531e9afc900fc5be92dfcc4be376b96f


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 0dc5b4d977a94acc520d13e41293ae17ffe65e9b
</p>

Akrion:
Based on the agreed pattern of "Composition over Configuration" I am little hesitant to agree with the `useLoader` config property. Seems to me this is similar to what we discussed before where for the sake of saving few lines of code we are adding a layer of abstraction/configuration.

I would be interested to hear what other think about this 🤔

stellarhoof:
I agree with @Akrion. In general, we should be trying to simplify injectTreeNode, not adding more features to it. Right now it's responsible for wrapping components in a loader, injecting (which is deprecated) a tree node and also creating one if it's not present. And yet the name is `injectTreeNode` :/

We can merge if this is critical but if it's not, I think it's a better use of time to rip out the loader from it. My suggestion was (and still is) to only wrap the components for which it actually makes sense. For example, in https://github.com/smartprocure/bid-search/issues/1278 wrapping both a component and its parent in loaders will look ugly since you end up with two loaders in the same place.

daedalus28:
Agreed - I think the right approach here is to break injectTreeNode down into separate components and then just compose the desired behavior together.

PR #270, opened 2019-07-24
Add 'view less' link to Facet

michsa:
Being able to increase the size of a facet but not decrease or reset it was getting annoying, so I threw this together:

Screen Shot 2019-07-24 at 6 00 29 AM

This also factors out the magic `10`s into `minSize` and `sizeIncrement` variables, and exposes  them as props on Facet so that it's possible to set them with mapNodeToProps or defaultProps if we ever want to 🙂

### DIFF
--- a/src/exampleTypes/Facet.js
+++ b/src/exampleTypes/Facet.js
@@ -6,16 +6,6 @@ import { exampleTypes } from 'contexture-client'
 import { Flex } from '../layout/Flex'
 import injectTreeNode from '../utils/injectTreeNode'
 
-/*
-_.max([
-  _.min([
-    ceilTens(node.context.cardinality), 
-    node.size
-  ]) - 10,
-  10
-])
-*/
-
 let ceilTens = _.partial(_.ceil.convert({fixed: false}), [_, -1])
 
 let CheckboxDefault = props => <input type="checkbox" {...props} />
@@ -186,7 +176,7 @@ let Facet = injectTreeNode(
                   <a
                     key="more"
                     onClick={() =>
-                      tree.mutate(node.path, { size: node.size + sizeIncrement })
+                      tree.mutate(node.path, { size: (node.size || minSize) + sizeIncrement })
                     }
                     style={{ cursor: 'pointer' }}
                   >
--- a/src/exampleTypes/Facet.js
+++ b/src/exampleTypes/Facet.js
@@ -1,12 +1,13 @@
 import React, { useState } from 'react'
 import _ from 'lodash/fp'
+import { ciel } from 'lodash'
 import F from 'futil-js'
 import { observer, Observer } from 'mobx-react'
 import { exampleTypes } from 'contexture-client'
 import { Flex } from '../layout/Flex'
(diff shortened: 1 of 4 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


Take a look at this commit to see what happened in detail: f325bd49fcc4faeeb3f68b936e5ab71b05b07e99


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 02876b912c02a940c9a767f2152a2bb58d67968a
</p>

src/exampleTypes/Facet.js
```diff
@@ -135,29 +139,57 @@ let Facet = injectTreeNode(
...
+              {_.flow(
+                _.compact,
+                F.intersperse(' — ')
+              )([
+                _.min([node.size, node.context.options.length]) > minSize && (
+                  <a
+                    key="less"
+                    onClick={() =>
+                      tree.mutate(node.path, {
+                        size: _.max([
```
michsa: this also works (but might be a bit much):

```js
_.flow(
  _.min,
  ceilTens,
  _.add(-10),
  _.concat([10]),
  _.max
)([node.context.cardinality, node.size])
```

src/exampleTypes/Facet.js
```diff
@@ -135,29 +139,57 @@ let Facet = injectTreeNode(
...
+                          _.min([
+                            node.size,
+                            ceilTens(node.context.options.length),
+                          ]) - sizeIncrement,
+                          minSize,
+                        ]),
```
michsa: this also works (but might be a bit much):

```js
_.flow(
  _.min,
  ceilTens,
  _.add(-10),
  _.concat([10]),
  _.max
)([node.context.cardinality, node.size])
```

src/exampleTypes/Facet.js
```diff
@@ -135,29 +140,57 @@ let Facet = injectTreeNode(
...
+              )([
+                _.min([node.size, node.context.options.length]) > minSize && (
+                  <a
+                    key="less"
+                    onClick={() =>
+                      tree.mutate(node.path, {
+                        size: _.max([
+                          _.min([
+                            node.size,
+                            ceilTens(node.context.options.length),
```
daedalus28: Just following up on our conversation - this is clever, but doesn't work if you've customized the interval.

Also, lets factor this out into a more reusable "clamp" thing since this logic might be usable in other places (like term stats)

daedalus28: something like this: https://runkit.com/daedalus28/clampmodel

daedalus28:
Closing since there is no real use case for this and it's gotten a bit stale. We can revisit later if need be.

PR #272, opened 2019-07-26
kill `injectTreeNode`

michsa:
Breaks injectTreeNode into `withNode`, `withAutoAdd`, and `withLoader` - naming scheme adjusted to match HOC convention, since we're not using `inject` anymore 🙂

Theoretically they're all composable together in any order, but there are some caveats:

1. Now that it accepts and handles the `path` prop, `withAutoAdd` effectively does everything that `withNode` does plus the add stuff, so it doesn't really make sense to compose the two in the first place. And the path stuff is so trivial that there's really not much reason _not_ to have it -- before, it was a configuration thing where `path` was used to fetch and `group` + `nodeKey` were used to add, but that's unnecessary now that they're separate functions.

2. As before, both `withNode` and `withAutoAdd` will throw if they can't produce a node, so if you _were_ going to compose the two for some reason, you'd have to put `withNode` first and give it the `allowEmptyNode` config prop.

   It's worth mentioning that adding a fourth piece to handle the check and throw  (`withEmptyNodeError`?) would let us clean that junk out of the other functions, at the cost of less informative error messages. I don't feel strongly about this one, so I'll leave it up to you guys to decide if it's worth it.

Stuff that works/is done:
- it actually functions perfectly as-is, thanks to the super lazy default export hack

Stuff that's broken/WIP:
- `withAutoAdd` is untested and definitely broken, at the very least on account of the contexture-client `add` return value thing. I still have to come up with a fresh test case for it, since I can't find anything that actually uses injectTreeNode's auto-add functionality ¯\\\_(ツ)\_/¯
- the super lazy default export hack is obviously a temporary measure and everything still has to be renamed (which is trivial but is seriously gonna clog up the diffs)
- removing loaders from the rest of the components that don't need them still needs to happen and may as well be part of this PR
- as usual, the names could probably stand to be improved


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


Take a look at this commit to see what happened in detail: 920dbb050561dc1e2de8f50ee921ca6f6e4b3182


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against f115d860ff8c417d4357c36ad21f432dffc1294f
</p>

src/FilterButtonList.js
```diff
@@ -73,50 +72,48 @@ let GroupBox = ({ nodeJoinColor, children, nested, className }) => (
   </Flex>
 )
 
-let FilterButtonList = InjectTreeNode(
-  observer(
```
michsa: this whole diff is just the indent change from removing `observer` (plus renaming `injectTreeNode`)

src/FilterButtonList.js
```diff
@@ -11,10 +10,10 @@ import {
 } from './layout'
 import DefaultIcon from './DefaultIcon'
 import DefaultMissingTypeComponent from './DefaultMissingTypeComponent'
-import InjectTreeNode from './utils/injectTreeNode'
+import { withNode, withLoader } from './utils/injectTreeNode'
```
michsa: I was using these components for testing anyway, so I figured I'd include the changes in the initial PR as a demo of what the new usage would look like

src/utils/injectTreeNode.js
```diff
@@ -1,57 +1,54 @@
...
+  
+  if (!node) {
+    if (!type)
+      throw Error(`Node not provided and cannot be autogenerated ('type' property is required)`)
+  
+    // now that client exposes `types` on the tree, we can look up the type object instead of having to pass it in
+    nodeProps = nodeProps || _.flow(_.get(['types', type, 'reactors']), _.keys)(tree)
+    let newNode = { type, key: nodeKey, ..._.pick(['field', ...nodeProps], props) }
+    // relies on client update to await dispatch and return new node on `tree.add`
+    node = tree.add(group, newNode)
```
michsa: this PR is probably going to depend on a contexture-client change no matter what, even if it's not a change to `add` like it's assuming here

src/exampleTypes/ResultCount.js
```diff
@@ -1,17 +1,9 @@
...
 
-let ResultCount = injectTreeNode(
-  observer(({ node, display = x => x }) =>
-    node.context.response.results.length
-      ? display(node.context.response.totalRecords)
-      : 'No Results'
-  ),
-  {
-    ...exampleTypes.results,
-    style: { display: 'inline-block' },
```
michsa: todo: test this

src/utils/hoc.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
+import { observer } from 'mobx-react'
+import _ from 'lodash/fp'
+import StripedLoader from './StripedLoader'
+
+export let withNode = (Component, { allowEmptyNode = false } = {}) => ({
+  tree,
+  node,
+  path,
+  ...props
```
daedalus28: Since we're passing along all of the props including tree, node, and path, it might be cleaner to just call this props, destructure inside the function, and just pass props along

Something like this:

```js
export let withNode = (Component, { allowEmptyNode = false } = {}) => props => {
  let { tree, node, path } = props
  props.node = node || (tree && path && tree.getNode(path))

  if (!props.node && !allowEmptyNode)
    throw Error(`Node not provided, and couldn't find node at ${path}`)

  return <Component {...props} />
}
```

src/utils/hoc.js
```diff
@@ -0,0 +1,42 @@
...
+  observer(({ Loader = StripedLoader, node, ...props }) => (
+    <Loader loading={node && node.updating} {...loaderProps}>
+      <Component node={node} {...props} />
+    </Loader>
+  ))
+
+export let contexturify = (Component, config, loaderProps) =>
+  _.flow(
+    observer,
+    c => withNode(c, config),
```
daedalus28: based on this composition, it kinda seems like maybe withNode and withLoader should have their args flipped and be curried 🤔

src/utils/hoc.js
```diff
@@ -0,0 +1,42 @@
...
+    c => withLoader(c, loaderProps)
+  )(Component)
+
+// this is used for the text components
+export let withTreeLens = Component => ({
+  tree,
```
daedalus28: it might be cleaner not to destructure out tree and node:

```js
export let withTreeLens = Component => ({ prop = 'value', ...props }) => (
  <Component {...{ lens: props.tree.lens(props.node.path, prop), ...props }} />
)
```

src/utils/hoc.js
```diff
@@ -0,0 +1,42 @@
...
+
+export let withNode = (Component, { allowEmptyNode = false } = {}) => ({
+  tree,
+  node,
+  path,
+  ...props
+}) => {
+  node = node || (tree && path && tree.getNode(path))
+
+  if (!node && !allowEmptyNode)
```
daedalus28: Is there any reason to support `allowEmptyNode`? it's the only reason for this HOC to have config and I don't see why this would ever be ok.

michsa: you're right, this was an oversight - I assumed it wasn't safe to take out (yet) because of the inject stuff in QueryBuilder/FilterList, but I just checked and it looks like it's fine even with that still in 👍

src/exampleTypes/ResultCount.js
```diff
@@ -1,18 +1,12 @@
...
     node.context.response.results.length
       ? display(node.context.response.totalRecords)
-      : 'No Results'
-  ),
-  {
-    ...exampleTypes.results,
-    style: { display: 'inline-block' },
-  }
+      : 'No Results',
+  {},
```
daedalus28: now that contexturify only takes 2 props, I think you meant to remove this

src/FilterButtonList.js
```diff
@@ -11,56 +10,58 @@ import {
...
-          )}
-          <div className="filter-component">
-            <Dynamic
-              Component={MissingTypeComponent}
-              tree={tree}
-              node={node}
-              path={_.toArray(node.path)}
-              {...mappedProps}
-            />
+let FilterButtonItem = withLoader(
```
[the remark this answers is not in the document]
michsa: Quick note for the record: it's not really an add; we're just moving the loader part of the composition from FilterButtonList to the individual FilterButtonItems instead, in order to stop the whole list from rerendering (and killing all its children's modals) when the group node is marked for update. I believe something similar needs to be done for the other search interfaces too.

src/themes/greyVest/index.js
```diff
@@ -923,19 +923,17 @@ export let Pager = props => (
...
-  observer(({ tree, node, ...props }) => (
-    <>
-      <ExampleTypes.ResultTable tree={tree} node={node} {...props} />
-      <Flex style={{ justifyContent: 'space-around', padding: '10px' }}>
-        <Pager tree={tree} node={node} />
-      </Flex>
-    </>
-  ))
-)
+export let PagedResultTable = contexturify(({ tree, node, ...props }) => (
```
stellarhoof: I don't think you need to wrap this one.

src/utils/hoc.js
```diff
@@ -0,0 +1,39 @@
...
+
+export let withLoader = Component =>
+  observer(({ Loader = StripedLoader, node, ...props }) => (
+    <Loader loading={node && node.updating}>
+      <Component node={node} {...props} />
+    </Loader>
+  ))
+
+// I am a band-aid, please rip me off as quickly as possible
+export let withInlineLoader = Component =>
```
stellarhoof: I'd say that it makes more sense to have a style prop be applied to the loader itself instead of the wrapped component. Then you could get rid of this hoc.

src/FilterAdder.js
```diff
@@ -21,4 +20,4 @@ let FilterAdder = ({ tree, node, path, fields, Picker, uniqueFields }) => {
   )
 }
 
-export default InjectTreeNode(observer(FilterAdder), { allowEmptyNode: true })
+export default contexturify(FilterAdder)
```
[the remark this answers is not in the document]
daedalus28: It is no longer needed - it only really made sense for the `addNode` case

src/utils/hoc.js
daedalus28: We did this intentionally to simplify the API of the HOC. The next change is going to remove this by allowing the loader to be overriden for the one component that uses it

src/themes/greyVest/index.js
daedalus28: I think you're right - Result Table and Pager are already wrapped

PR #280, opened 2019-08-02
TagsInput is now smarter with input

Akrion:
Related: https://github.com/smartprocure/bid-search/issues/3024

Issue: Entering strings separated by ‘,’ was broken. Users could enter ‘,’ and it would result in two empty tags, or enter ‘a’, after which ‘a,a’ which would result in 3 ‘a’ in the tags input.

image

image

New logic fixes that and also filters and adds only the new unique and new tags. So if you have ‘a,b,c’ and you enter ‘d,a,b,c’ only ‘d’ would be added etc.

Tags entry:

TagsInput1

Tags Copy/Paste:

TagsInput2

### DIFF
--- a/src/layout/TagsInput.js
+++ b/src/layout/TagsInput.js
@@ -10,14 +10,18 @@ import OutsideClickHandler from 'react-outside-click-handler'
 
 let isValidInput = (tag, tags) => !_.isEmpty(tag) && !_.includes(tag, tags)
 
-let getUniqueTagsFromInput = (tag, tags, splitCommas = true) => {
+// extracts only the unique (and not already used tags) from the user tags input
+// e.g. if "x,y,z" are already entered then only "d" would be added if the user tris to enter "x,y,d"
+// returns a string with either single tag or multiple unique tags separated by comma
+let getUniqueTagsStringFromInput = (tag, tags, useCommaSeparator = true) => {
   let trimmedTag = _.trim(tag)
-  if (splitCommas && _.includes(',', trimmedTag)) {
-    // Extract only the unique and not present in the current tags collection tags from the string
+  if (useCommaSeparator && _.includes(',', trimmedTag)) {
+    // Extract only the unique (and not present in the current tags collection) tags from the string
     let newUniqTags = _.difference(
       _.uniq(_.compact(_.split(',', trimmedTag))),
       tags
     )
+    // if not empty return the joined string which would be split to separate tags in `addTag` if splitCommas is set
     return _.isEmpty(newUniqTags) ? null : newUniqTags.join(',')
   }
   return trimmedTag
@@ -155,7 +159,7 @@ let TagsInput = withState('state', 'setState', () =>
                   state.currentInput = e.target.value
                 }}
                 onBlur={() => {
-                  let input = getUniqueTagsFromInput(
+                  let input = getUniqueTagsStringFromInput(
                     state.currentInput,
                     tags,
                     splitCommas
@@ -166,7 +170,7 @@ let TagsInput = withState('state', 'setState', () =>
                   }
                 }}
                 onKeyDown={e => {
-                  let input = getUniqueTagsFromInput(
+                  let input = getUniqueTagsStringFromInput(
                     state.currentInput,
                     tags,
                     splitCommas
--- a/src/layout/TagsInput.js
+++ b/src/layout/TagsInput.js
@@ -81,14 +81,18 @@ let TagsInput = withState('state', 'setState', () =>
     }) => {
       let containerRef
       let inputRef
-      if (splitCommas)
-        addTag = _.flow(
-          _.split(','),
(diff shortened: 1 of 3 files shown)


src/layout/TagsInput.js
```diff
@@ -8,9 +8,16 @@ import { Flex } from './Flex'
...
-let isValidTag = (tag, tags) => {
-  let cleanTag = _.trim(tag)
-  return !_.isEmpty(cleanTag) && !_.includes(cleanTag, tags)
+let isValidInput = (tag, tags) => !_.isEmpty(tag) && !_.includes(tag, tags)
+
+let getUniqueTagsFromInput = (tag, tags, splitCommas=true) => {
+  let trimmedTag = _.trim(tag)
+  if(splitCommas && _.includes(',', trimmedTag)) {
+    // Extract only the unique and not present in the current tags collection tags from the string
+    let newUniqTags = _.difference(_.uniq(_.compact(_.split(',', trimmedTag))), tags)
```
Akrion: `_.flow` was too long and had issues with `_.difference`- had to use `_.filter` etc so it ended up less concise

decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


Take a look at this commit to see what happened in detail: a7bd40638442d6e55211b383a632a5720d1adf5f


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against f31e7ea8a01d8e3d6085afab194f53e4f557852f
</p>

src/layout/TagsInput.js
```diff
@@ -8,9 +8,19 @@ import { Flex } from './Flex'
...
+
+let getUniqueTagsFromInput = (tag, tags, splitCommas = true) => {
+  let trimmedTag = _.trim(tag)
+  if (splitCommas && _.includes(',', trimmedTag)) {
+    // Extract only the unique and not present in the current tags collection tags from the string
+    let newUniqTags = _.difference(
+      _.uniq(_.compact(_.split(',', trimmedTag))),
+      tags
+    )
+    return _.isEmpty(newUniqTags) ? null : newUniqTags.join(',')
```
daedalus28: This is a little weird at least from a naming perspective - we're saying that if `splitCommas` is true, we're still returning a string joined by a comma 🤔

Akrion: Added comments and renamed the function.

src/layout/TagsInput.js
```diff
@@ -8,9 +8,23 @@ import { Flex } from './Flex'
...
+
+// extracts only the unique (and not already used tags) from the user tags input
+// e.g. if "x,y,z" are already entered then only "d" would be added if the user tris to enter "x,y,d"
+// returns a string with either single tag or multiple unique tags separated by comma
+let getUniqueTagsStringFromInput = (tag, tags, useCommaSeparator = true) => {
+  let trimmedTag = _.trim(tag)
+  if (useCommaSeparator && _.includes(',', trimmedTag)) {
+    // Extract only the unique (and not present in the current tags collection) tags from the string
+    let newUniqTags = _.difference(
+      _.uniq(_.compact(_.split(',', trimmedTag))),
```
michsa: Re. your comment about `flow` breaking with `difference`, this seems to work (though I only tested it briefly, so I might have missed something):

```es
_.flow(
  _.split(','),
  _.compact,
  _.uniq,
  _.difference(_, tags)
)(trimmedTag)
```

or you could keep `difference` outside and just flow `split`, `compact` and `uniq`:

```es
_.difference(
  _.flow(
    _.split(','),
    _.compact,
    _.uniq,
  )(trimmedTag)
)(tags)
```

I personally think these are cleaner than the nested calls, but it's up to you 🙂

Akrion: Right.

I usually would go with `flow` but in this case this is not as straight forward to follow with the `_.difference(_, tags)` or with the `_difference` and then the inside `_.flow`. In this case I prefer the nested approach.

src/layout/TagsInput.js
daedalus28: I think the issue is that, in the case of `splitCommas`, we're taking a string, splitting it, deduping it, joining it, splitting it, iterating, and then adding. Let's just improve the logic on line 102 where that's happening so we don't do extra work

Akrion:
Cleanup done.

Akrion:
No there is no issue with the `splitCommas` scenario, but yes in the case of single tags where `splitCommas` is falsy. I handled that case as well now.

PR #288, opened 2019-08-08
Theme context API & demo stories

michsa:
For this PR, I ended up going with a super-simple but (hopefully) clear and informative set of stories to demonstrate the API, rather than eg. reproducing an actual search interface like FilterList using themeContext. The goal is for this to go into the storybook as-is and function as documentation.

Example:

Screen Shot 2019-08-09 at 11 17 22 AM

### DIFF
--- a/src/utils/theme.js
+++ b/src/utils/theme.js
@@ -3,26 +3,31 @@ import _ from 'lodash/fp'
 
 let ThemeContext = React.createContext()
 
-export let ThemeProvider = ({ children, theme }) =>
-  <ThemeContext.Provider value={theme}>
-    {children}
-  </ThemeContext.Provider>
-
-export let mergeNestedTheme = (key = '', theme) => _.flow(
-  _.pickBy((val, k) => _.startsWith(`${key}.`, k)),
-  _.mapKeys(x => _.replace(`${key}.`, '', x)),
-  _.defaults(theme)
-)(theme || React.useContext(ThemeContext))
-
-export let mergeNestedThemePath = (path, theme) => _.reduce(
-  (acc, key) => mergeNestedTheme(key, acc), 
-  theme || React.useContext(ThemeContext), 
-  path || []
+export let ThemeProvider = ({ children, theme }) => (
+  <ThemeContext.Provider value={theme}>{children}</ThemeContext.Provider>
 )
 
-export let ThemeConsumer = ({ children, path }) => children(mergeNestedThemePath(path))
+export let mergeNestedTheme = (key = '', theme) =>
+  _.flow(
+    _.pickBy((val, k) => _.startsWith(`${key}.`, k)),
+    _.mapKeys(x => _.replace(`${key}.`, '', x)),
+    _.defaults(theme)
+  )(theme || React.useContext(ThemeContext))
 
-export let withTheme = (defaults = {}, name) => Component => ({ theme: propTheme, ...props }) => {
+export let mergeNestedThemePath = (path, theme) =>
+  _.reduce(
+    (acc, key) => mergeNestedTheme(key, acc),
+    theme || React.useContext(ThemeContext),
+    path || []
+  )
+
+export let ThemeConsumer = ({ children, path }) =>
+  children(mergeNestedThemePath(path))
+
+export let withTheme = (defaults = {}, name) => Component => ({
+  theme: propTheme,
+  ...props
+}) => {
   name = name || Component.displayName || Component.name
   let contextTheme = mergeNestedTheme(name, React.useContext(ThemeContext))
 
--- a/src/utils/theme.js
+++ b/src/utils/theme.js
@@ -1,0 +1,36 @@
+import React from 'react'
+import _ from 'lodash/fp'
+
+let ThemeContext = React.createContext()
+
+export let ThemeProvider = ({ children, theme }) =>
+  <ThemeContext.Provider value={theme}>
+    {children}
+  </ThemeContext.Provider>
+
+export let mergeNestedTheme = (key = '', theme) => _.flow(
+  _.pickBy((val, k) => _.startsWith(`${key}.`, k)),
+  _.mapKeys(x => _.replace(`${key}.`, '', x)),
+  _.defaults(theme)
+)(theme || React.useContext(ThemeContext))
+
+export let mergeNestedThemePath = (path, theme) => _.reduce(
+  (acc, key) => mergeNestedTheme(key, acc), 
+  theme || React.useContext(ThemeContext), 
+  path || []
+)
+
+export let ThemeConsumer = ({ children, path }) => children(mergeNestedThemePath(path))
+
+export let withTheme = (defaults = {}, name) => Component => ({ theme: propTheme, ...props }) => {
+  name = name || Component.displayName || Component.name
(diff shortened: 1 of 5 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Your PR has lint warnings. Please consider fixing these.
      
<details>
  <summary>theme.js</summary>
  <p><a href="https://github.com/smartprocure/contexture-react/blob/feature/theme-context/src/utils/theme.js" target="_blank">src/utils/theme.js</a></p>
  
  <strong>Unexpected console statement.</strong>
  <code>Line 21: console.log('component name', Component.displayName) // this will not be available much of the time</code>
</details>
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


Take a look at this commit to see what happened in detail: f1721bd2d05112a995ee2cfb63eed016fd06a7a6


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against de91c18857d772bab56dc0b4ef13aff541b7e24c
</p>

src/utils/theme.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react'
+import _ from 'lodash/fp'
+
+let ThemeContext = React.createContext()
+
+export let ThemeProvider = ({ children, theme }) => (
```
daedalus28: This probably doesn’t add much value over just exporting ThemeContext.Provider directly. I’m fine with calling the prop `value` instead of `theme`

src/utils/theme.js
```diff
@@ -0,0 +1,41 @@
...
+  <ThemeContext.Provider value={theme}>{children}</ThemeContext.Provider>
+)
+
+export let mergeNestedTheme = (key = '', theme) =>
+  _.flow(
```
daedalus28: Let’s only merge if there are any overrides (especially since it’s less common that there will be any)

src/utils/theme.js
```diff
@@ -0,0 +1,41 @@
...
+  )
+
+export let ThemeConsumer = ({ children, path }) =>
+  children(mergeNestedThemePath(path))
+
+export let withTheme = (defaults = {}, name) => Component => ({
+  theme: propTheme,
+  ...props
+}) => {
+  name = name || Component.displayName || Component.name
```
daedalus28: I think it might be better to force explicit opt in (so that you can only override things that have passed their name into withTheme). This will do two things: 1. If there is no name, skip over all the merging stuff and just treat it like regular context, and 2. Allow us to not worry about getting component names exactly right (otherwise tweaking display names would actually become breaking changes)

src/utils/theme.js
```diff
@@ -0,0 +1,41 @@
...
+  _.reduce(
+    (acc, key) => mergeNestedTheme(key, acc),
+    theme || React.useContext(ThemeContext),
+    path || []
+  )
+
+export let ThemeConsumer = ({ children, path }) =>
+  children(mergeNestedThemePath(path))
+
+export let withTheme = (defaults = {}, name) => Component => ({
```
daedalus28: Not sure we need defaults at the withTheme level

src/utils/theme.js
michsa: For what it's worth, I left this in for two reasons:
- parity with ThemeConsumer (which is kind of an extra anyway)
- it matches the theme APIs for styled-components and emotion

Both of those are pretty minor, so no objection here 🙂

daedalus28: The advantage of not doing is that it's react context, nothing special about our API. Plus, its less code

src/utils/theme.js
```diff
@@ -0,0 +1,42 @@
...
+    hasNested(key),
+    _.flow(
+      _.pickBy((val, k) => _.startsWith(`${key}.`, k)),
+      _.mapKeys(x => _.replace(`${key}.`, '', x)),
+      _.defaults(theme)
+    )
+  )(theme)
+
+export let mergeNestedThemePath = (theme, path) =>
+  _.reduce((acc, key) => mergeNestedTheme(acc, key), theme, path || [])
```
daedalus28: I think this could be reduced to just

```js
export let mergeNestedThemePath = _.reduce(mergeNestedTheme)
```

src/utils/theme.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react'
+import _ from 'lodash/fp'
+import F from 'futil-js'
+
+export let ThemeContext = React.createContext()
+
+let hasNested = _.curry(
+  (key, theme) =>
+    key && theme && F.findIndexed((v, k) => _.startsWith(`${key}.`, k), theme)
```
daedalus28: findIndex should handle empty/null things, so you shouldn't need to check `theme`, and we shouldn't even enter in to this branch if there is no `key`. Since this is curried anyway and only used here, you could simplify this to:

```js
let hasNested = key => F.findIndexed((v, k) => _.startsWith(`${key}.`, k))

File: src/utils/theme.js [hunk 7]
```diff
@@ -0,0 +1,42 @@
...
+  <ThemeContext.Consumer>
+    {theme => children(mergeNestedThemePath(theme, path))}
+  </ThemeContext.Consumer>
+)
+
+export let withTheme = name => Component => ({
+  theme: propTheme,
+  ...props
+}) => {
+  let contextTheme = mergeNestedTheme(React.useContext(ThemeContext), name)
```
daedalus28: If there is no `name`, let's just not even do the merge thing so we're not cloning unnecessarily.

src/utils/theme.js
```diff
@@ -0,0 +1,42 @@
...
+  let newTheme = _.merge(contextTheme, propTheme)
```
daedalus28: same as above - no need to do a merge if we only have one or the other

src/utils/theme.js
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    <ThemeContext.Provider value={newTheme}>
```
daedalus28: Seems like we probably also shouldn't wrap in provider unless we're sure we need a new context, although I'm not sure if really adds any overhead

daedalus28: I just realized - this will also merge the prop theme onto the context for children, which means component theme stuff will cascade down to children, too. I wonder if that's what we want 🤔

michsa: I'm not sure how heavy Provider is, but I think we should just assume it's fine until we have cause to optimize 🙂

michsa: It was intentional to make the last story work (merging nested theme props) - so tentatively yes

PR #291, opened 2019-08-13
Theme refactor: Blueberry

michsa:
1. Layout:
   - **Update layout components to use the `theme` prop and `withTheme` HOC**, as necessary. (The majority of layout components don't have theme props, and the ones that do are pretty basic, which makes this relatively small and low-risk)
   - **Export a basic theme object from `layout`** containing all the basic components in addition to the components themselves
2. Blueberry theme:
   - **Split `blueberry.js`** into a directory with an index and component files (for parity with GreyVest, and for improved readability in general)
   - **Export a theme object from `blueberry`** containing layout components and example types, in addition to the components themselves
   - **Add a new IMDB demo story** that uses the Blueberry theme with ThemeContext to render themed components, **alongside the existing story** that uses the exported components.

Basically, we now have the "Blueberry Theme" story and an identical-looking "Blueberry Theme with Theme API" story. This is the plan for GreyVest as well -- export all the same components, including re-exporting exampleTypes, so existing implementations can migrate to the new API over time.


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>:exclamation: This PR is BIG (+710 -387)  
Please keep it below 500 net changes</td>
    </tr>
  
<tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


Take a look at this commit to see what happened in detail: 6c0c1193b9779922acba07ef1a181e7a419ef9a7


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 882812c15a8532eb3ab36f45e6fe8836c1d10193
</p>

src/themes/blueberry/index.js
```diff
@@ -0,0 +1,60 @@
...
+  Button,
+  Input,
+  Checkbox,
+  RadioList: ButtonRadio,
+  'FieldPicker.Item': ListGroupItem,
+  'TagsInput.TagComponent': TagComponent,
+  Pager,
+}
+
+let { TypeMap, ...ExampleTypes } = ExampleTypeConstructor(baseTheme)
```
michsa: Note that the example types are still using regular component props without any new theme stuff - this works because `ExampleTypeConstructor` (the `src/exampleTypes` default export) effectively takes a theme object as its argument anyway. 

`exampleTypes` still needs the same treatment as `layout`, but it wasn't necessary for this PR

src/layout/index.js
```diff
@@ -1,39 +1,65 @@
...
+  CheckButton,
+  Dynamic,
+  ExpandableTable,
   Flex,
   Grid,
+  LensInput,
   Modal,
+  ModalPicker,
+  NestedPicker,
+  // FieldPicker: NestedPicker,
```
michsa: This is evidence of a bug where nested themes don't work when the name a component is given by `withTheme` (or `ThemeConsumer`) is different from its key in the theme.

Uncomment this to see the Blueberry picker break, even though `FieldPicker` component still defaults to `NestedPicker`.

src/themes/blueberry/index.js
```diff
@@ -0,0 +1,62 @@
...
+  'Adder.Item': ListGroupItem,
+  Wrap,
+  Fonts,
+  Style,
+  Table,
+  Button,
+  Input,
+  Checkbox,
+  RadioList: ButtonRadio,
+  'FieldPicker.Item': ListGroupItem,
```
michsa: This is evidence of a bug where nested themes don't work when the name a component is given by `withTheme` (or `ThemeConsumer`) is different from its key in the theme.

If the line in `src/layout/index.js` is uncommented (`FieldPicker` is set to `NestedPicker` in the default layout theme), this override won't be merged, because `withTheme('NestedPicker')` does not recognize keys starting with `'FieldPicker'`.

stellarhoof:
What if we set the default components in an exported default theme as opposed to having defaults in all the props?

daedalus28:
> What if we set the default components in an exported default theme as opposed to having defaults in all the props?

This is what I thought we'd do as well. Having defaults at the component level allows us to defend against incomplete themes, but I'm not sure that matters. It would likely be substantially simpler to just put a default theme - we'd do at the point of `React.createContext`, so I'm pretty sure it would even work if there was no `Provider` at all.

src/themes/blueberry/index.js
daedalus28: Not sure I follow here - this is by design 🤔 Things that can overriden in a theme should have a single, canonical name.

src/layout/index.js
```diff
@@ -1,39 +1,65 @@
...
+import Popover from './Popover'
 import Portal from './Portal'
-import CheckButton from './CheckButton'
+import SpacedList from './SpacedList'
 import StepsAccordion, { AccordionStep } from './StepsAccordion'
+import { Tag, TagsInput } from './TagsInput'
+import TextHighlight from './TextHighlight'
 
-export {
+export let theme = {
```
daedalus28: whats with the double export here?

src/themes/blueberry/index.js
```diff
@@ -0,0 +1,62 @@
...
+  Button,
+  ButtonRadio,
+  ListGroupItem,
+  Wrap,
+  TagComponent,
+  PagerItem,
+  Adder,
+}
+
+export let theme = {
```
daedalus28: it seems like "theme" should just be the default export

src/layout/LensInput.js
```diff
@@ -1,10 +1,14 @@
...
-))
-LensInput.displayName = 'LensInput'
+let LensInput = ({ lens, theme: { Input = 'input' }, ...props }) => (
+  <Input {...F.domLens.value(lens)} {...props} />
+)
 
-export default LensInput
+export default _.flow(
+  observer,
+  withTheme('LensInput')
```
daedalus28: Let's be conservative with adding these names. There's no reason I can think of where we'd need to have a different "input" just for LensInput. We can always add these later, but once they're in, changing or removing them is a breaking change

src/themes/blueberry/index.js
michsa: yep, that would've been smart 😅

michsa:
The last commits added a global default theme made up of basic layout components, and removed a bunch of the default props now that they're no longer necessary. It works, and it does reduce code, but there are some caveats:

1. Any components which are themselves themed cannot be imported into `theme` as part of the global default, or they'll cause circular dependencies, so we still have to include component-level defaults for them. That include most example types, as well as these layout components: https://github.com/smartprocure/contexture-react/blob/766c7e1ae84d0d34e077494ea361f4e26438a30c/src/layout/index.js#L23-L30

2. The nice thing about putting the defaults directly on context is that we don't need a ThemeProvider to access them; `ThemeConsumer`/`withTheme` alone are enough.

   However, that means they're *always passed in* -- and because component-level defaults are just default props, they're overridden by the globals. As a result, theme-wrapped components are now dependent on ThemeContext to override the global defaults, which means they can no longer work as self-contained exports.

   Here's Blueberry's ButtonRadio to illustrate. With the Blueberry theme in context, it correctly uses the Blueberry Button component; but when exported individually, its Button prop becomes the default layout button:

   Screen Shot 2019-08-13 at 7 09 37 PM
   Screen Shot 2019-08-13 at 7 09 55 PM

stellarhoof:
> theme-wrapped components are now dependent on ThemeContext to override the global defaults, which means they can no longer work as self-contained exports.

I'm confused. Components using `withTheme` will be wrapped with their own `ThemeProvider`. Why wouldn't they work as standalone exports?

src/layout/index.js
```diff
@@ -1,39 +1,81 @@
...
+import RemoveTagIcon from './RemoveTagIcon'
+import SpacedList from './SpacedList'
 import StepsAccordion, { AccordionStep } from './StepsAccordion'
+import { Tag, TagsInput } from './TagsInput'
+import TextHighlight from './TextHighlight'
 
-export {
+import { defaultTheme } from '../utils/theme'
+
+let theme = {
```
stellarhoof: I don't think these components should be part of the themeable: Awaiter, Flex, Grid, Dynamic, LensInput, ModalPicker, PopoverContents, Portal, StepsAccordion, TagComponent, Wrap.

stellarhoof: I don't think it makes sense for these components to be themeable: Awaiter, Flex, Grid, Dynamic, LensInput, ModalPicker, PopoverContents, Portal, StepsAccordion, TagComponent, Wrap.

src/layout/index.js
```diff
@@ -1,39 +1,81 @@
...
+  StepsAccordion,
+  Table: 'table',
   Tag,
+  TagComponent: Tag,
   TagsInput,
-  LensInput,
+  TextHighlight,
+  Wrap: 'i',
+}
+F.mergeOn(defaultTheme, theme)
```
stellarhoof: Why don't you like `require`ing the base theme here: https://github.com/smartprocure/contexture-react/pull/291/files#diff-e1c6d5d34fa2955dc0df3284a194bce3R27

daedalus28: From @michsa
> it still doesn’t work on context initialization, though, so we’d lose providerless defaults

stellarhoof: I'm not sure that'd be the case. Every themeable component is wrapped in a `ThemeProvider` which merges the base theme so initializing the context is unnecessary.

daedalus28: Initializing the context is necessary to support usage without a `ThemeProvider`

src/layout/index.js
michsa: Yep, some of those are explicitly not used by any theme components and shouldn't have been there in the first place (Flex, Grid, Portal). I'll also take out Awaiter, Dynamic, StepsAccordion & AccordionStep. The rest are definitely being used somewhere or other - I'll leave them in for now, and if the usage turns out to be trivial, we can take them out later 🙂

src/utils/theme.js
```diff
@@ -20,17 +23,18 @@ export let mergeNestedTheme = (theme, key) =>
   )(theme)
 
 let useTheme = (name, propTheme) =>
-  mergeOrReturn(
+  mergeOrReturnAll([
+    defaultTheme,
```
daedalus28: We should not merge this in here. React will provide this by default if there is nothing on `Provider` - this would just defend against incomplete themes which we don't care about. Even worse, this will make so that we'll always be doing a merge on every render of every component because the default and context themes will most often always be present

src/layout/index.js
stellarhoof: Ok as per @daedalus28 clarification, we don't want to merge the default theme on `useTheme` every time, so this comment is irrelevant.

src/utils/futil.js
```diff
@@ -13,6 +13,12 @@ export let FlattenTreeLeaves = Tree =>
 export let PlainObjectTree = F.tree(onlyWhen(_.isPlainObject))
 export let flattenPlainObject = F.whenExists(FlattenTreeLeaves(PlainObjectTree))
 
+let canMerge = a => (!_.isEmpty(a) ? a : false)
```
daedalus28: slightly more compact:

```js
let canMerge = a => !_.isEmpty(a) && a
```

src/themes/blueberry/index.js
```diff
@@ -0,0 +1,61 @@
...
+import TagComponent from './TagComponent'
+import PagerItem from './PagerItem'
+import Adder from './Adder'
+
+export let Pager = props => (
+  <ExampleTypes.ResultPager Item={PagerItem} {...props} />
+)
+
+let baseTheme = {
+  ...defaultTheme,
```
daedalus28: You went through the trouble of allowing partial themes, so no need to spread on the default theme here

src/themes/blueberry/index.js
```diff
@@ -0,0 +1,61 @@
...
+  Adder,
+  Wrap,
+  Fonts,
+  Style,
+  Table,
+  Button,
+  Input,
+  Checkbox,
+  RadioList: ButtonRadio,
+  'TagsInput.TagComponent': TagComponent,
```
daedalus28: Why does this need to be scoped? What else uses TagComponent?

src/themes/blueberry/index.js
```diff
@@ -0,0 +1,61 @@
...
+import Adder from './Adder'
+
+export let Pager = props => (
+  <ExampleTypes.ResultPager Item={PagerItem} {...props} />
+)
+
+let baseTheme = {
+  ...defaultTheme,
+  Adder,
+  Wrap,
```
daedalus28: This should be renamed to be clearer - e.g. HighlightWrap or something

src/themes/blueberry/index.js
```diff
@@ -0,0 +1,61 @@
...
+import PagerItem from './PagerItem'
+import Adder from './Adder'
+
+export let Pager = props => (
+  <ExampleTypes.ResultPager Item={PagerItem} {...props} />
+)
+
+let baseTheme = {
+  ...defaultTheme,
+  Adder,
```
daedalus28: Adder is also a bit unclear

src/themes/blueberry/index.js
```diff
@@ -0,0 +1,61 @@
...
+  Fonts,
+  Style,
+  Table,
+  Button,
+  Input,
+  Checkbox,
+  RadioList: ButtonRadio,
+  'TagsInput.TagComponent': TagComponent,
+  Pager,
+  Item: ListGroupItem,
```
daedalus28: Item should definitely get renamed

src/utils/futil.js
michsa: 🤦‍♀

src/themes/blueberry/index.js
michsa: you're right - nothing does, so it doesn't 👍

src/FilterAdder.js
```diff
@@ -2,22 +2,35 @@ import _ from 'lodash/fp'
...
 
-let FilterAdder = ({ tree, node, path, fields, Picker, uniqueFields }) => {
+let FilterAdder = ({
+  tree,
+  node,
+  path,
+  fields,
+  theme: { AdderPicker },
+  uniqueFields,
+  label = 'Add Custom Filter',
```
michsa: With the addition of an `AdderPicker` theme key, and with ModalPicker as the default `AdderPicker`, the only thing left for [ModalFilterAdder](https://github.com/smartprocure/contexture-react/blob/1d1e96a72611c175550f0093b20f9c56a5876b1a/src/ModalFilterAdder.js) to do was to pass a `label` prop into its AdderPicker component. (A regular picker component like NestedPicker has no need for a label, but ModalPicker uses it as the text for the button to open the modal.)

So, let's kill it completely*, and just let the base FilterAdder handle it instead 🙂

<sub>* ModalFilterAdder is still being used in several places that have yet to be refactored, and replacing the internals with this version breaks stuff, so best to hold off on actually killing it for now</sub>

src/themes/blueberry/index.js
michsa: ListItem it is 🙂

src/FilterAdder.js
stellarhoof: +1 on this. Man your comments are getting so professional. Love the footnote.

stellarhoof: Why not `FilterAdder.Picker`? Isn't this the whole point of having nesting in the themes?

src/layout/RemoveTagIcon.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+
+let RemoveTagIcon = props => (
+  <span className="tags-input-tag-remove" {...props}>
```
daedalus28: since this is only used by the default theme, I think maybe this should just be part of the default tag component and not be on the theme

daedalus28:
Merging so this doesn't get stale

PR #298, opened 2019-08-20
THEME DOC

michsa:
Let's take another shot at the whole "document the theme API" thing, except (hopefully) make it actually useful this time

The main point of interest here is the list of default theme components. It's meant to be totally comprehensive - the source of truth for what should and should *not* be in the GreyVest theme (ie, stuff that just lives in and is exportable from the GreyVest component library).

Which would make this PR a convenient place for those sorts of discussions, I think 🙂

Marked as WIP because there are still a couple of open questions regarding certain keys -- mostly having to do with whether specific usages of some generic thing like List or Adder should have their own top-level keys (possible theme bloat), or just be nested (cognitive load and other complications). The documentation itself also deserves to be expanded a bit, though I believe the priority on that is pretty low.

### DIFF
--- a/docs/theme.md
+++ b/docs/theme.md
@@ -151,10 +151,8 @@ This section documents all of the **theme components** used by contexture-react 
 | `DateInput` | `value`, `onChange` | A generic date entry component | Date |
 | `NumberInput` | same as basic `input` | An input of type `number` | Geo, Number |
 | `TagsInput` | `tags`, `addTag`, `removeTag`, `submit`, `tagStyle`, `placeholder`, `splitCommas`, `style`, `onBlur`, `onInputChange`, `onTagClick` | A text input field that turns input into `Tag`s | TagsQuery, TagsText |
-| `PopoverTagsInput` | same as TagsInput, plus `PopoverContents` | A TagsInput that opens a popover when a tag is clicked. Uses the `TagsInput` theme component. | TagsQuery, TagsText |
 | `Tag` | `value`, `removeTag`, `tagStyle`, `onClick` | A tag component with a button to remove the tag. Used in TagsInput. | none |
-| `Picker` | `options`, `onChange` | Renders a list of selectable options | ResultTable |
-| `ModalPicker` | `options`, `onChange`, `label` | A picker inside a modal, with a button to open it. Uses `Picker`, `Button`, and `Modal` theme components. | none |
+| `NestedPicker` | `options`, `onChange` | Renders a list of selectable options | ResultTable |
 
 ### Containers
 
@@ -178,7 +176,7 @@ This section documents all of the **theme components** used by contexture-react 
 | --- | --- | --- |
 | `Globals` | `children` | This component is rendered inside `ThemeProvider`, and wraps its children. For use with other providers and/or standalone globals like stylesheets. |
 | `BarChart` | `height`, `borderColor`, `min`, `max` | Used in the DateHistogram and TermsStats example types |
-| `MissingTypeComponent` | `node` | Used as a fallback component in search interfaces when no other component is found for a node's contexture type |
+| `UnmappedNodeComponent` | `node` | Used as a fallback component in search interfaces for nodes that are not mapped to any other component |
 
 
 ## Theme Authoring
--- a/docs/theme.md
+++ b/docs/theme.md
@@ -1,0 +1,147 @@
+# Theme API
+
+A theme in contexture-react is simply an object -- which we'll call a **theme object** for clarity -- that maps **theme keys** to React components. The theme object lives inside contexture-react's `ThemeContext`, which uses React context under the hood.
+
+Since they are really just identifiers for components, theme keys should follow the same naming conventions as React components themselves. For the purpose of consuming the API, we'll refer to a theme key together with the component it identifies as a **theme component**.
+
+## Using the Theme API
+
+To use **theme components** inside your own component definitions, your component must pull them from the theme object within contexture-react's `ThemeContext`.
+
+There are two ways to get the theme from context: the `ThemeConsumer` component, and the `withTheme` function.
+
+### ThemeConsumer
+
+The `ThemeConsumer` helper component takes a render function as its child, and passes the **theme object** from context into that function. This example demonstrates how to create a basic IconButton component using the `Icon` and `Button` **theme components**, plus `ThemeConsumer`:
+
+```jsx
+// IconButton.js
+
+import React from 'react'
+import { ThemeConsumer } from 'contexture-react/src/utils/theme'
+
+let IconButton = ({ icon, children, ...props }) => (
+  <ThemeConsumer>
+    {theme =>
+      <theme.Button {...props}>
+        <theme.Icon icon={icon} />
+        {children}
+      </theme.Button>
+    }
+  </ThemeConsumer>
+)
+export default IconButton
+```
+
+### withTheme
+
+The `withTheme` higher-order component injects the theme object from context into its component argument as the `theme` prop. Here is the same IconButton example using `withTheme`:
+
+```jsx
+// IconButton.js
+
+import React from 'react'
+import { withTheme } from 'contexture-react/src/utils/theme'
+
+let IconButton = ({ theme, icon, children, ...props }) => (
+  <theme.Button {...props}>
+    <theme.Icon icon={icon} />
+    {children}
+  </theme.Button>
+)
+export default withTheme(IconButton)
+```
+
+`ThemeConsumer` and `withTheme` both accomplish the same thing in different ways. You're free to use whichever one you like best. 🙂
+
+
+## Default theme components
+
+The default theme in contexture-react is GreyVest. The full list of **theme components** included in GreyVest is shown in the table below.
+
+| Key | Expected props | Notes |
+| --- | --- | --- |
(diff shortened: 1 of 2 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against e969060127818f91cc37e406de509679e29f3a4a
</p>

docs/theme.md
```diff
@@ -0,0 +1,138 @@
...
+| `AdderPicker` | `options`, `onChange`, `label` | *this key is actually based on specific usage in the FilterAdder component, and should probably just be removed from the defaults in favor of a nested style* |
+| `BarChart` | `height`, `borderColor`, `min`, `max` | Used in the DateHistogram and TermsStats example types |
+| `Box` | `children` | A generic container element |
+| `Button` | `isActive`, `primary`, `onClick`, `children` | A generic button |
+| `Checkbox` | `checked`, `onChange` | A generic checkbox |
+| `CheckButton` | `checked`, `onClick`, `children` | A button with an `onClick` handler and an on/off state, containing a checkbox that displays this state *(this really should be extrapolated to some kind of generic BinaryStateButton)* |
+| `ExpandableTable` | `data`, `columns`, `recordKey`, `sortField`, `sortDir`, `columnSort` | Used in the TermsStatsTable and CheckableTermsStatsTable example types |
+| `Icon` | `icon`, `onClick` | A generic icon component |
+| `Fonts` | none | For holding `<link>` elements to fonts used in the theme |
+| `Input` | same as basic `input` | A generic input component |
```
daedalus28: Does anything actually use this directly? With example types, we started to have explicit NumberInput, TextInput, etc.

There are also definitely a bunch of missing things like `Select`, `Radio`, etc

docs/theme.md
```diff
@@ -0,0 +1,138 @@
...
+| `ListItem` | `children` | A generic list item |
+| `MissingTypeComponent` | `node` | Used as a fallback component in search interfaces when no other component is found for a node's contexture type |
+| `Modal` | `isOpen`, `children` | A generic modal component |
+| `PagerItem` | `active`, `disabled` | A list item for Pager components *(make this nested instead?)* |
+| `Picker` | `options`, `onChange` | Renders a list of selectable options |
+| `PickerItem` | `active`, `disabled` | A list item for Picker components *(make this nested instead?)* |
```
daedalus28: It occurs to me that if PickerItem is only every used by Picker (as I imagine it is), perhaps it should just be an intrinsic part of `Picker`. Does it really make sense to override PickerItem but not Picker, or Picker but not PickerItem?

docs/theme.md
```diff
@@ -0,0 +1,138 @@
...
+| `Checkbox` | `checked`, `onChange` | A generic checkbox |
+| `CheckButton` | `checked`, `onClick`, `children` | A button with an `onClick` handler and an on/off state, containing a checkbox that displays this state *(this really should be extrapolated to some kind of generic BinaryStateButton)* |
+| `ExpandableTable` | `data`, `columns`, `recordKey`, `sortField`, `sortDir`, `columnSort` | Used in the TermsStatsTable and CheckableTermsStatsTable example types |
+| `Icon` | `icon`, `onClick` | A generic icon component |
+| `Fonts` | none | For holding `<link>` elements to fonts used in the theme |
+| `Input` | same as basic `input` | A generic input component |
+| `ListItem` | `children` | A generic list item |
+| `MissingTypeComponent` | `node` | Used as a fallback component in search interfaces when no other component is found for a node's contexture type |
+| `Modal` | `isOpen`, `children` | A generic modal component |
+| `PagerItem` | `active`, `disabled` | A list item for Pager components *(make this nested instead?)* |
```
daedalus28: Just like I mentioned on PickerItem - this seems like maybe it should just be part of "pager"

docs/theme.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+`ThemeConsumer` and `withTheme` both accomplish the same thing in different ways. You're free to use whichever one you like best. 🙂
+
+
+## Default theme components
+
+The default theme in contexture-react is GreyVest. The full list of **theme components** included in GreyVest is shown in the table below.
+
+| Key | Expected props | Notes |
```
daedalus28: I don't think organizing this table alphabetically is all that helpful - let's try to group them into logical sections (so that, for example, ExpandableTable and Table are next to each other). I could see us having a section for inputs (checkbox, input, radio list, tags input, etc), containers (modal, popover, box), and maybe another section. Things that are hard to clearly put into one section or another are definitely things that should be scrutinized

docs/theme.md
michsa: Yep, but it's really the same as TextInput - best to keep it consistent with the others, so Input should get renamed.

Select was definitely needed. I don't believe anything uses a Radio component though - pretty sure it's just RadioList

docs/theme.md
michsa: > Does it really make sense to override PickerItem but not Picker, or Picker but not PickerItem?

I think it actually does - in fact, the blueberry theme overrides only PickerItem (to be the same as its regular ListItem component), and doesn't touch Picker itself. Overriding the picker component lets you add functionality (eg modals and nesting), while the item component is more of a styling thing, so they're different concerns.

(As a side note, this is basically how it was with TextHighlight and HighlightWrap; I nixed the `HighlightWrap` key anyway because they're so tiny that it's trivial to declare both at once.)

With `Picker.ListItem` rather than `PickerItem`, you'd only have to override the `ListItem` key in a case like blueberry's, rather than having to remember to do `PickerItem` (and whatever `OtherItem`) too. It's nice, but is it worth making sure that every picker is named `Picker` for withTheme?

michsa:
I figured out what GreyVest's ListGroupItem is - it's for the popovers in ResultsTable that have icons next to each item. The extra styles add a 5px grid-gap between the icon and the text:

Screen Shot 2019-08-20 at 10 15 50 PM

But, this is also causing the overflow issue in the other popovers, which don't have icons:

Screen Shot 2019-08-20 at 10 23 31 PM

So we have a couple of options. We could:
1. Just stick with ListItem everywhere, and use an alternative for the icon styling that doesn't break iconless ListItems
2. give ResultTable a theme name and use a nested theme for its ListItems, or
3. create a new theme component, IconListItem, and have ResultTable use that instead.

It's a small thing for sure, but this definitely seems like the sort of decision that will help us pin down a more general philosophy about how to approach cases like this, so I thought it was worth discussing 🙂

daedalus28:
> I figured out what GreyVest's ListGroupItem is - it's for the popovers in ResultsTable that have icons next to each item.

This is probably a misuse of the list group stuff. There's a missing "drop down" thing which should probably be there (which doesn't necessarily make sense to be the same as a generic "List Group" thing)

michsa:
Now that ThemeProvider renders style globals, how do we want to approach simple style overrides for general layout components like StepsAccordion (ie, those that are library exports only, not part of the base theme)?

I see two ways:

1. We could continue with what we were doing before the refactor, which is re-exporting the layout component with styles attached (classNames or vanilla JSS for now, emotion eventually)
2. We could ensure the components have canonical classNames, eg `steps-accordion`, and just add CSS overrides for that className to the theme's stylesheet

On the consumer's side, the main difference between the two is where to import the component from - the theme component library (eg src/greyVest) in the first case, or the general library (src/layout) in the second case. 

Option 2 is arguably more flexible, since it allows the client app to swap out themes without having to change any imports. (This isn't that important in the case of our product, but as precedent for the library, it seems valuable to me.) It would also create less overhead from re-exported components.

I bring this up because we were definitely leaning toward option 1 for everything, and doing away with the Style component completely. On consideration, though, I think the CSS route might be the way to go in this specific case (that is, pure styling overrides for non-theme library components, which happens a few times in GreyVest).

stellarhoof:
With the release of the storybook docs we could also import this in our stories!

docs/theme.md
```diff
@@ -0,0 +1,293 @@
...
+| `DropdownItem` | `children` | A generic list item used in dropdown menus | ResultTable |
+| `PagerItem` | `active`, `disabled` | A list item for Pager components | ResultPager |
+| `TextHighlight` | `pattern`, `text`, `Wrap` | Renders the text given in the `text` prop, with the parts that match the `pattern` prop wrapped in the `Wrap` component | none |
+| `Table` | same as basic `table` | A generic table component | ResultTable |
+
+### Other
+
+| Key | Expected props | Notes |
+| --- | --- | --- |
+| `Globals` | `children` | This component is rendered inside `ThemeProvider`, and wraps its children. For use with other providers and/or standalone globals like stylesheets. |
```
daedalus28: Based on the other PR, lets call this `Root`

PR #303, opened 2019-08-27
Add Material UI theme

michsa:
A demo thing to put the theme API through its paces.

Lessons:

- Converting between our **lens API** and Material's `{ open, onClose }` API was the *least* painful part of the whole thing, since it can trivially be handled by a wrapper function (`lensify` in `themes/material/utils.js`). It looks like supporting a non-lens API for our components won't be necessary after all 🙂

- Material doesn't supply **dynamic multi-option form components** like our RadioList and Select, so those had to be reimplemented using the Material base components. We could consider allowing the base components (eg RadioButton and SelectItem) to be themed as well - it would have saved some code in this particular situation, though I'm not sure it's worth it in general.

- Material's **Popover** component takes an `anchorEl` prop that expects a reference to a DOM element. I still like our self-anchoring popovers, but this might be nice to have as an option.

- Our **TagsInput** component does a lot of *stuff* that doesn't come standard with generic implementations -- comma splitting, opening a popover when you click a tag, dynamically styling the tags, submitting on double-enter, etc. It's easy to get a third-party TagsInput working, but reimplementing our extra stuff on top of someone else's API (assuming we care about that) becomes painful quickly.

  If we do want to factor out some of that functionality, the popover stuff is our best bet. All the TagsInput implementations I looked at had some way to provide a tag click callback (either directly, or through a "tag props" prop). We could add that to ours and then do something similar to Picker and ModalPicker, with a PopoverTagsInput and a vanilla one.

  (As a side note, all of the other implementations also had huge APIs with a ton of variation (eg `onAdd` vs `handleAddition` vs `onChange`, and that's just the basics) so there's no point trying to standardize there.)

- Material's **Table** API has components for `<thead>` and `<tbody>`:
  ```jsx
  <Table> {/* <-- our Table */}
    <TableHead>
      <TableRow>
        <TableCell /> {/* <-- our TableHeaderCell */}
      </TableRow>
    </TableHead>
    <TableBody>
      <TableRow> {/* <-- our TableRow or Row prop */}
        <TableCell /> {/* <-- our TableCell or Cell prop */}
      </TableRow>
    </TableBody>
  </Table>
  ```
  We can definitely clean up some of the pure layout stuff from the ResultTable/TermsStatsTable example types, and at that point we may as well follow their example.

- Material has two primary ways to override styles: a ThemeProvider, and a `classes` prop that works on all their components and accepts a CSS-in-JS tree. The takeaway here is that we should have a Globals theme key that can be used for external providers/wrappers as well as styles and fonts, so we can initialize them inside ours.

Action items:

1. `Globals` theme component
2. Clean up table stuff
3. Maybe split TagsInput into TagsInput and PopoverTagsInput
4. Maybe enhance our popover with an optional anchor prop
5. Maybe add theme support for individual RadioButtons and SelectItems


stellarhoof:
I'm getting lots of errors/warnings in the console when browsing the storybook. We should minimize those.

src/utils/react.js
```diff
@@ -12,3 +13,21 @@ export let wrapDisplayName = (name, Component) => Wrapped => {
...
   return Wrapped
 }
+
+// these are for futil :)
+// (x -> y) -> string -> {x} -> y
+let getWith = _.curry((customizer, path, object) =>
+  customizer(_.get(path, object))
+)
+// ({a} -> {b}) -> {a} -> {a, b}
+let expandObject = _.curry(transform => obj => ({ ...obj, ...transform(obj) }))
```
daedalus28: For the curry to work, transform and obj should be on the same function

src/themes/material/Select.js
```diff
@@ -0,0 +1,29 @@
...
+import _ from 'lodash/fp'
+import { Select as MaterialSelect, MenuItem } from '@material-ui/core'
+
+let Select = ({
+  options,
+  placeholder = 'Please Select...',
+  value = '',
+  ...props
+}) => (
+  <Observer>
```
daedalus28: Is this needed, or can we use the HOC?

src/themes/material/Style.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import { CssBaseline } from '@material-ui/core'
+
+let Style = () => (
+  <>
+    <style>{`
+      th span {
```
daedalus28: This seems like a very generic selector if we assume there are other tables in an app using material

src/themes/material/Select.js
michsa: we can use the HOC! https://github.com/mobxjs/mobx-react/blob/f9763b/test/stateless.test.js#L86

src/utils/react.js
michsa: 🤦‍♀🤦‍♀🤦‍♀🤦‍♀🤦‍♀

src/themes/material/Style.js
michsa: just to note - it would be nice to use material's styling API for this, but as far as I know the `th span` inside the material table isn't accessible via its custom selector stuff (since it's actually part of ResultTable).

emotion will make things nicer, but for now I think the solution is just to stick a className on there and style it in the root like we do with grey vest.

src/utils/react.js
```diff
@@ -12,3 +13,21 @@ export let wrapDisplayName = (name, Component) => Wrapped => {
...
+let expandObject = _.curry((transform, obj) => ({ ...obj, ...transform(obj) }))
+
+// (a -> {b}, k) -> Component<{k: a}> -> Component<{b}>
+export let lensify = _.flow(
+  getWith,
+  expandObject,
+  mapProps
+)
+
+let open = (...lens) => ({ open: F.view(...lens), onClose: F.off(...lens) })
```
stellarhoof: This is not generic. It's specific to material-ui

src/themes/material/Select.js
```diff
@@ -0,0 +1,28 @@
...
+import { Select as MaterialSelect, MenuItem } from '@material-ui/core'
+
+let Select = ({
+  options,
+  placeholder = 'Please Select...',
+  value = '',
+  ...props
+}) => (
+  <MaterialSelect fullWidth displayEmpty value={value} {...props}>
+    {placeholder && <MenuItem value="">{placeholder}</MenuItem>}
```
stellarhoof: You probably want their `displayEmpty` prop for the placeholder.

src/themes/material/Tag.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import F from 'futil-js'
+import { Chip } from '@material-ui/core'
+
+let Tag = ({ removeTag, value, tagStyle, ...props }) => (
+  <Chip
+    onDelete={() => removeTag(value)}
+    label={value}
+    style={F.callOrReturn(tagStyle, value)}
```
stellarhoof: You'll override `style` if passed on `props` below

src/utils/react.js
```diff
@@ -12,3 +13,21 @@ export let wrapDisplayName = (name, Component) => Wrapped => {
   Wrapped.displayName = `${name}(${getDisplayName(Component)})`
   return Wrapped
 }
+
+// these are for futil :)
+// (x -> y) -> string -> {x} -> y
+let getWith = _.curry((customizer, path, object) =>
```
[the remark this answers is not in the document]
daedalus28: This is already going to be a thing :)

https://github.com/lodash/lodash/issues/4458

src/utils/react.js
```diff
@@ -12,3 +13,21 @@ export let wrapDisplayName = (name, Component) => Wrapped => {
...
+  customizer(_.get(path, object))
+)
+// ({a} -> {b}) -> {a} -> {a, b}
+let expandObject = _.curry((transform, obj) => ({ ...obj, ...transform(obj) }))
```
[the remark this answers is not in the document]
daedalus28: This is also destined for futil and just lives here for now.

src/utils/react.js
daedalus28: This follows the same model as the futil domLens functions like `value`, `hover`, and `focus`:

https://github.com/smartprocure/futil-js/blob/master/src/lens.js#L83-L87

But you are right - it doesn't seem generic enough to live in futil itself, which is why it lives here.

src/themes/material/Select.js
michsa: isn't `displayEmpty` a boolean? It's already on there in the line above

src/utils/react.js
stellarhoof: It's not generic because `open` and `onClose` are material UI props

src/themes/material/Select.js
stellarhoof: My bad. I meant `renderValue`

src/utils/react.js
daedalus28: thats fair, we should move open + openify to utils in material 👍

stellarhoof:
> The openify implementation is too abstract. Just make it a single HOC

So after talking with Sam, this seems to fall on the subjective side, no need to do anything here for now.

PR #307, opened 2019-08-30
Feature/add table class to table components

Akrion:
In regards to https://github.com/smartprocure/bid-search/issues/3225

Adds a simple `tableClassName` to the table component so we can add alternating class in bid-search.


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 09dee77423b999d43eaa979e0c969109b407e369
</p>

daedalus28:
This is already supported by the `Table` component which is generally how we change styling from contexture-react. In the app, you can just target gv-table though, right?

Akrion:
The requirement is to add the alternating background to only few screens. So adding this allows us to only add that prop to only the grids that need it. Class will be in bid-search. Otherwise adding a className to a ResultTable does nothing

daedalus28:
@Akrion you can still do that by passing an alternate `Table` component via the props on a per instance basis. That covers everything - whether you want to add a classname, inline style, or even use a totally different dom element

daedalus28:
@Akrion this exact use case is covered in our storybook:

https://github.com/smartprocure/contexture-react/blob/master/stories/exampleTypes/ResultTable.stories.js#L13-L32

stellarhoof:
@daedalus28 Passing styles/className to a component is pretty standard. I don't see the harm here. It's a bit of an overkill to override the table component just to put a className on it.

src/exampleTypes/ResultTable.js
```diff
@@ -337,6 +337,7 @@ let ResultTable = InjectTreeNode(
       mapNodeToProps = () => ({}),
       Icon = DefaultIcon,
       Row = Tr,
+      tableClassName,
```
stellarhoof: Actually can we rename this to just `className`

Akrion:
My reaction as well @stellarhoof when I realized that we have to overwrite the `Table` tag via function instead of simply providing a className to the component.

daedalus28:
The intention of being able to change the table tag is to handle these use cases - otherwise, we'd add className, style, and whatever else. It starts to get really messy though, because you'd eventually want to have table className/style, row className/style, etc - so instead of that, we just let you customize the table rendering by taking control of it.

This idea is also a bit clearer after the v2 branch is in because it makes a separation between contexture-react (which generally is not very concerned with styling/theming), Grey Vest the component library (which is 100% about styling and theming), and the Grey Vest contexture-react theme which stitches them together.

@michsa should chime in here as well since long term the idea was that we'd do the actual styling with emotion (for the component library).

daedalus28:
Closing as this has gotten stale and we haven't needed it. We can revisit in the future if it comes up again (after the style guide project)

PR #309, opened 2019-09-05
Fix more v2 conflicts

michsa:
My IDE isn't handling this merge well for some reason, so let's try it this way


michsa:
Never mind, I did this completely wrong 😅

PR #331, opened 2019-09-21
Fix Modal/Popover

stellarhoof:
- We need to observe `F.view`, which is being used in `openBinding`
- `openBinding` should not fail if there's no `open` lens used
- Added stories to use both API styles to make sure they work

### DIFF
--- a/src/greyVest/Modal.js
+++ b/src/greyVest/Modal.js
@@ -2,42 +2,44 @@ import React from 'react'
 import { observer } from 'mobx-react'
 import Portal from './Portal'
 import { openBinding } from './utils'
-import { expandProp } from '../utils/react'
 
-let Modal = ({ isOpen, onClose, children, style = {}, className = '' }) => (
-  <Portal>
-    {isOpen && (
-      <div
-        style={{
-          position: 'fixed',
-          top: 0,
-          bottom: 0,
-          left: 0,
-          right: 0,
-          backgroundColor: 'rgba(0,0,0,0.3)',
-          padding: 50,
-          overflowY: 'auto',
-          zIndex: 1000,
-          display: 'flex',
-          justifyContent: 'space-around',
-          alignItems: 'flex-start',
-        }}
-        onClick={onClose}
-        className={`default-modal-bg ${className}`}
-      >
+let Modal = ({ open, children, style = {}, className = '' }) => {
+  let { isOpen, onClose } = openBinding(open)
+  return (
+    <Portal>
+      {isOpen && (
         <div
           style={{
-            backgroundColor: '#fff',
-            ...style,
+            position: 'fixed',
+            top: 0,
+            bottom: 0,
+            left: 0,
+            right: 0,
+            backgroundColor: 'rgba(0,0,0,0.3)',
+            padding: 50,
+            overflowY: 'auto',
+            zIndex: 1000,
+            display: 'flex',
+            justifyContent: 'space-around',
+            alignItems: 'flex-start',
           }}
-          onClick={e => e.stopPropagation()}
-          className="default-modal-wrap"
+          onClick={onClose}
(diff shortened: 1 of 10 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 7c329be58f641c1e6930de9f5ad1c9ff02a07bc1
</p>

src/greyVest/Modal.js
```diff
@@ -2,42 +2,44 @@ import React from 'react'
...
-          zIndex: 1000,
-          display: 'flex',
-          justifyContent: 'space-around',
-          alignItems: 'flex-start',
-        }}
-        onClick={onClose}
-        className={`default-modal-bg ${className}`}
-      >
+let Modal = ({ open, children, style = {}, className = '' }) => {
+  let { isOpen, onClose } = openBinding(open)
```
daedalus28: This breaks the new API and makes it so you can’t pass isOpen and onClose directly.

src/greyVest/Popover.js
```diff
@@ -2,29 +2,32 @@ import React from 'react'
...
-            fontWeight: 'normal',
-            textAlign: 'left',
-            background: 'white',
-            border: '1px solid #ebebeb',
-            ...style,
-          }}
-        >
-          {children}
+let Popover = ({ open, children, style }) => {
+  let { isOpen, onClose } = openBinding(open)
```
daedalus28: Same as above, this breaks the API

src/greyVest/utils.js
```diff
@@ -1,6 +1,7 @@
 import F from 'futil-js'
 
-export let openBinding = (...lens) => ({
-  isOpen: F.view(...lens),
-  onClose: F.off(...lens),
-})
+export let openBinding = lens =>
```
michsa: We still need the spread here to support all the futil lens types, so with no args `lens` would be `[undefined]` instead of `undefined`. Maybe something like

```js
export let openBinding = (...lens) => 
  !_.isEmpty(lens) && _.head(lens) && {
    isOpen: F.view(...lens),
    onClose: F.off(...lens),
  }
```

michsa: We still need the spread here to support all the futil lens types, so with no args `lens` would be `[undefined]` instead of `undefined`. Maybe something like

```js
export let openBinding = (...lens) => 
  !_.isEmpty(lens) && _.head(lens) && {
    isOpen: F.view(...lens),
    onClose: F.off(...lens),
  }
```

stellarhoof: Ok just to summarize from our conversation: The current code fails to accoutn for lens being an array, but an array is not a valid lens right now. The lens prop API for view is `F.view(attr, obj)`, not `F.view([attr, obj])`, so unless we want to account for array lenses, this should be fine

stellarhoof: There's no way to pass two args as one so `lens` won't be `(prop, obj)`

src/utils/react.js
```diff
@@ -17,6 +17,6 @@ export let wrapDisplayName = (name, Component) => Wrapped => {
 
 // (k, a -> {b}) -> Component<{k: a}> -> Component<{b}>
 export let expandProp = _.flow(
-  expandObjectBy,
+  (key, fn) => expandObjectBy(key, F.whenExists(fn)),
```
daedalus28: doesnt this take 3 args?

PR #335, opened 2019-09-30
Add FacetSelect

michsa:
It's a facet, but smaller!

facet2


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against d5c8186194e85e9467d71437c72662aeb3c2012f
</p>

src/exampleTypes/FacetSelect.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import F from 'futil-js'
+import _ from 'lodash/fp'
+import { contexturify } from '../utils/hoc'
+import Async from 'react-select/lib/Async'
+
+let getOptions = node =>
+  _.map(({ name }) => F.autoLabelOption(name), node.context.options)
```
daedalus28: Auto label isn’t right here, but we can support a display function like in facet

daedalus28: Also don’t forget to show counts and cardinality somewhere since facet gets it

michsa: counts and display functions are in 👍

Regarding cardinality -- as far as I can tell, there's no way to customize how many results react-select will show at a time, which means no "view more" functionality on FacetSelect. Of course there's nothing stopping us from adding cardinality anyway, but it'd just kinda be sitting there by itself, and I dunno if it's useful enough on its own to justify the space it would take up. (Personally, I'm inclined to say it's not -- the best thing about FacetSelect is how compact it is)

daedalus28: View more/less would still matter because it’d affect how many options are actually loaded. It’s not a react select customization - it would control the facet node via contexture - react select will just render al the options we return, so we control the size effectively that way. Cardinality (and maybe even the view more control) could be done with the group rendering thing from the first example in customizing built ins: https://react-select.com/advanced#replacing-builtins

michsa:
Just realized I never posted this:

Screen Shot 2019-10-14 at 6 25 26 PM

I'd thought the number of results was a hardcoded react-select thing, but duh, it was just the default result count for facet the whole time.

Also worth noting - I added a couple of extra hide options to both facet components. From the changelog:
  * `modeToggle`: hide Facet's mode selector (the "Include"/"Exclude" switch) 
  * `cardinality`: hide "Showing X of Y" and the "view more" link

michsa:
I reverted the added hide options and removed the item counts on tags that weren't updating, so this should be ready for final review 🙂

PR #337, opened 2019-10-03
Add GridItem component and Grid 'row' prop

michsa:
GridItem is a div with props for most of the basic grid-positioning-related things you'd want to do with a grid item.

@daedalus28, @stellarhoof - what are your thoughts on `colX` vs `columnX` (eg `colSpan`/`columnSpan`) in the prop API for this one? I like the brevity of `col`, but our Grid component uses `column`, so we might want to stick with it out of consistency. Either way, we probably want to weigh the options on this one before merging.

### DIFF
--- a/src/greyVest/Grid.js
+++ b/src/greyVest/Grid.js
@@ -1,12 +1,16 @@
 import React from 'react'
 
-let Grid = ({ style, columns, gap, ...props }) => (
+let Grid = ({ style, columns, rows, gap, ...props }) => (
   <div
     style={{
       display: 'grid',
       ...(columns && {
         gridTemplateColumns: columns,
         msGridTemplateColumns: columns,
+      }),
+      ...(rows && {
+        gridTemplateRows: rows,
+        msGridTemplateRows: rows,
       }),
       ...(gap && { gridGap: gap }),
       ...style,
--- a/src/greyVest/GridItem.js
+++ b/src/greyVest/GridItem.js
@@ -1,0 +1,30 @@
+import React from 'react'
+
+let GridItem = ({
+  column,
+  colStart,
+  colEnd,
+  colSpan,
+  row,
+  rowStart,
+  rowSpan,
+  rowEnd,
+  children,
+  style,
+}) => (
+  <div
+    style={{
+      gridColumn: column,
+      gridColumnStart: colStart,
+      gridColumnEnd: colEnd || (!!colSpan && `span ${colSpan}`),
+      gridRow: row,
+      gridRowStart: rowStart,
+      gridRowEnd: rowEnd || (!!rowSpan && `span ${rowSpan}`),
+      ...style,
+    }}
+  >
+    {children}
+  </div>
+)
+
+export default GridItem
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 2.2.0
+* GreyVest: Add GridItem component
+* GreyVest: Add `row` prop to Grid
+
 # 2.1.1
 * Add `search-layout` and `search-layout-{mode}` classNames to SearchLayout
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "contexture-react",
-  "version": "2.1.0",
+  "version": "2.2.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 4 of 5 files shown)


src/greyVest/Grid.js
```diff
@@ -1,13 +1,17 @@
...
-let Grid = ({ style, columns, gap, ...props }) => (
+let Grid = ({ style, columns, rows, gap, ...props }) => (
   <div
     style={{
       display: 'grid',
       ...(columns && {
         gridTemplateColumns: columns,
         msGridTemplateColumns: columns,
       }),
+      ...(rows && {
```
stellarhoof: :+1: I had this in my bucket list at some point. Nice!

src/greyVest/GridItem.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react'
+
+let GridItem = ({
```
stellarhoof: I'm not thrilled about the start,end,span props. I personally like to drop down to css for this with `row` and `column` as you have it here. It encourages actually learning grid

stellarhoof:
@michsa I'm not sure how far you're willing to go with this PR but I like this API: https://styled-css-grid.js.org/

src/greyVest/GridItem.js
michsa: You got me - I just didn't want to actually learn grid myself 😂

Yeah, I don't know what I was thinking with `span`. I'm still inclined toward keeping `start` and `end` though, even if they're not strictly necessary, since they directly mirror a commonly-used part of the CSS API.

daedalus28:
> @michsa I'm not sure how far you're willing to go with this PR but I like some of the ideas on this API: https://styled-css-grid.js.org/

That thing looks pretty nice. Makes thing maybe we should just use it.

`middle` and `center` as props is awesome

michsa:
> That thing looks pretty nice.  Makes me think maybe we should just use it.

Dang, it really does. I agree - let's use it!

FYI, I'm not sure whether this part counts as a breaking change with our current Grid:

> When the **columns** prop is a number, it is a shorthand for **grid-template-columns: repeat(N, 1fr)**.

This is technically a behavior change on passing a number to `columns`, which currently creates a grid with one column at that pixel width. Of course, I doubt anybody would actually use it that way (and we definitely don't), so we can probably get away with saying that the current behavior is unsupported. @daedalus28, thoughts?

stellarhoof:
So that library requires styled-components as a peer dep so if we use it we've got to buy into it. I'm 100% ok with that for the record :dancer:

daedalus28:
> So that library requires styled-components as a peer dep so if we use it we've got to buy into it. I'm 100% ok with that for the record 💃 . It's just that we'd have to stop exporting Grid from here. It'd be pretty weird if a user of the library was required to install styled-components just to use a simple layout helper.

I have mixed `emotion`s about this 😂

daedalus28:
> > That thing looks pretty nice.  Makes me think maybe we should just use it.
> 
> Dang, it really does. I agree - let's use it!
> 
> FYI, I'm not sure whether this part counts as a breaking change with our current Grid:
> 
> > When the **columns** prop is a number, it is a shorthand for **grid-template-columns: repeat(N, 1fr)**.
> 
> This is technically a behavior change on passing a number to `columns`, which currently creates a grid with one column at that pixel width. Of course, I doubt anybody would actually use it that way (and we definitely don't), so we can probably get away with saying that the current behavior is unsupported. @daedalus28, thoughts?

That's fine because we didn't document that behavior, so it wasn't "officially" part of our public API

stellarhoof:
Ok then let's just copy their implementation. Their components are just prop wrappers so it's trivial

src/greyVest/Grid.js
```diff
@@ -1,14 +1,35 @@
...
+  inline = false,
+  style,
+  ...props
+}) => (
+  <Component
     style={{
-      display: 'grid',
-      ...(columns && {
-        gridTemplateColumns: columns,
-        msGridTemplateColumns: columns,
```
daedalus28: If we remove this, we will lose whatever basic IE grid support we have. I’m fine with that, but we should at least note it in the changelog

src/greyVest/GridItem.js
daedalus28: Since we are copying the API of another component, we should include a link referencing it. A comment is sufficient

stellarhoof: Actually we just got inspired by it. After going over it, we didn't copy most of their API

PR #342, opened 2019-10-15
Feature/theme loader

michsa:
Makes the `withLoader` HOC (used in contexturify, formerly injectTreeNode, etc) look in theme for a Loader component if one isn't explicitly passed in.

Also fixes a really dumb bug where `withNode` and `withLoader` were in the wrong order, so `withLoader` couldn't see the updating status of nodes that were fetched in `withNode`, which was causing a bunch of loaders not to render. 

This included some loaders we *don't* want -- ResultPager, FilterList, and QueryBuilder all had their loaders reappear after the fix. The loaders on these top-level components were always unnecessary and were waiting on 2.0 to be removed - this PR finally finishes the job 🙂

### DIFF
--- a/src/utils/hoc.js
+++ b/src/utils/hoc.js
@@ -20,7 +20,8 @@ export let withLoader = Component =>
   _.flow(
     wrapDisplayName('withLoader', Component),
     observer
-  )(({ Loader, theme = {}, node, ...props }) => {
+  )(({ Loader, ...props }) => {
+    let { theme = {}, node } = props
     Loader = Loader || theme.Loader || StripedLoader
     return (
       <Loader loading={node && node.updating}>
@@ -34,17 +35,24 @@ export let withInlineLoader = Component =>
   _.flow(
     wrapDisplayName('withInlineLoader', Component),
     observer
-  )(({ Loader = StripedLoader, node, ...props }) => (
-    <Loader loading={node && node.updating} style={{ display: 'inline-block' }}>
-      <Component node={node} {...props} />
-    </Loader>
-  ))
+  )(({ Loader, ...props }) => {
+    let { theme = {}, node } = props
+    Loader = Loader || theme.Loader || StripedLoader
+    return (
+      <Loader
+        loading={node && node.updating}
+        style={{ display: 'inline-block' }}
+      >
+        <Component {...props} />
+      </Loader>
+    )
+  })
 
 export let contexturify = _.flow(
   observer,
   withNode,
-  withTheme,
-  withLoader
+  withLoader,
+  withTheme
 )
 
 // this is used for the text components
--- a/src/utils/hoc.js
+++ b/src/utils/hoc.js
@@ -50,8 +50,8 @@ export let withInlineLoader = Component =>
 
 export let contexturify = _.flow(
   observer,
-  withNode,
   withLoader,
+  withNode,
   withTheme
 )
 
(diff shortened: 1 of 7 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 6c8f2a31bc732d96207e0ce941feb7a4b894c22d
</p>

src/utils/hoc.js
```diff
@@ -20,27 +20,38 @@ export let withLoader = Component =>
...
+        <Component {...props} />
+      </Loader>
+    )
+  })
 
 export let contexturify = _.flow(
   observer,
-  withNode,
   withLoader,
+  withNode,
```
dubiousdavid: I don't understand this change. Why wouldn't you want `withNode` before `withLoader`?

src/utils/hoc.js
```diff
@@ -20,27 +20,38 @@ export let withLoader = Component =>
...
     wrapDisplayName('withLoader', Component),
     observer
-  )(({ Loader = StripedLoader, node, ...props }) => (
-    <Loader loading={node && node.updating}>
-      <Component node={node} {...props} />
-    </Loader>
-  ))
+  )(({ Loader, ...props }) => {
+    let { theme = {}, node } = props
+    Loader = Loader || theme.Loader || StripedLoader
```
stellarhoof: You're assuming there's a `theme` being passed down, which is not a good assumption. Just wrap this with `withTheme`. Also don't pass the `theme` down in the props to the Component since it can do so itself.

src/utils/hoc.js
stellarhoof: You're right, it should go before, which is what this change is doing. The HOCs here are being applied from innermost to outermost, so it'd be: `withTheme(withNode(withLoader(observer(Component)))`

dubiousdavid: I see. Makes sense. Thanks.

src/utils/hoc.js
dubiousdavid: Wouldn't it make sense to assume that if `theme` is a prop on the component that `withTheme` has already been used higher up? I don't see any cases where we have `withLoader` without `withTheme` preceding it.

stellarhoof: Perhaps right now we don't, but making assumptions about the calling environment increases coupling and causes subtle bugs.

dubiousdavid: By that logic, shouldn't this also be wrapped with `withNode`? What I am getting at is that dependencies are not necessarily a bad thing, and are actually already assumed.

stellarhoof: You're right I hadn't realized the node dependency. In that case we should add `withNode` to the mix here as well. It's just not good UX imo to have a `withLoader` that in order to use, you need to remember to also wrap it with `withNode` and `withTheme`. If it always needs to be done, then might as well export it with those two HOCs already applied.

dubiousdavid: I understand what you are saying, but I don't agree. I think we can revisit this later, but for now, I think this is the most consistent with the current state of the codebase.

Issue #347, opened 2019-10-29
Add moment to dependencies


michsa:
It's used in GreyVest's `DateInput`, so it's required for both the contexture-react and grey-vest packages.

UltimaBGD:
I can work on this if you would like.

UltimaBGD:
Is there a particular version in mind?

michsa:
Hey @UltimaBGD, welcome! 🤗

To answer your question, it looks like we're already using the [latest version](https://www.npmjs.com/package/moment) of moment, 2.2.4:

https://github.com/smartprocure/contexture-react/blob/317b7bff16b4bf9e9cda27828599cf767d86592f/package.json#L108

So the same version should be fine, the package just needs to be moved into the `dependencies` section instead.

UltimaBGD:
Gotcha, will have a PR up shortly.

michsa:
Closed by #348 🎉

PR #348, opened 2019-10-29
Add moment to dependencies

UltimaBGD:
This pull request references #347 and adds the latest version of moment to the dependencies list.


daedalus28:
@UltimaBGD thank you! We just need to also bump the version in the package.json and add an entry in the changelog. This should be a fix revision following semver, so we should bump to `2.4.1`

UltimaBGD:
Understood!

daedalus28:
This is great, thank you!

PR #354, opened 2019-11-05
Add logic to get results from the search bar action menu clicks

Akrion:
Added `actionWrapper` which when passed to the `SearchBar/TagsQuerySearchBar` allows for a feedback of the `ActionMenu` actions via the `after` and `onError` aspects.


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
      <td>

<details>
  <summary>ActionsMenu.js</summary>
  <p><a href="https://github.com/smartprocure/contexture-react/blob/feature/ActionBarFeedbackFn/src/exampleTypes/TagsQuery/ActionsMenu.js" target="_blank">src/exampleTypes/TagsQuery/ActionsMenu.js</a></p>
  
  <strong>'F' is defined but never used.</strong>
  <code>Line 3: import F from 'futil-js'</code>
</details>
  </td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against c263187ed3d25981c57178c7b6c2fdc0a7ce1d73
</p>

decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 506314ae8d476434f6e6fedeb098fbea9b49a452
</p>

stellarhoof:
Why not just display an error under the action?

Akrion:
Usually the common way to do this is the actual action menu closes and you get feedback via the `toast`. I would think that is the expected these days UI feedback. I do not think that the error should be on the action menu ... it is kind of weird. It is not a form per-se etc.

stellarhoof:
> Usually the common way to do this is the actual action menu closes and you get feedback via the toast

Or that. We've got toasts, errors under the component or even a commandbutton replica with a console error. They all seem better to me than this. I mean, there are so many actions performed everywhere that don't have callbacks, nor they should because it'd increase the complexity of the component so much. Just think resulttable uff

Akrion:
Well that was the intention ... this now allows for the main UI to show `toasts` based on the argument passed in the function `onActionMenuClick`. We would not want to couple and have an opinion about toast when it comes to `contexture-react` wold be my guess right?

Not set on this but we need something to give us feedback that action menu was successful or not and use the consumer of contexture-react to provide the user with the feedback.

stellarhoof:
Yeah toasts are opinionated for sure, which is why my 2c is to just show the error under the action.

Akrion:
Yeah not a fan of that ... from UI stand point just feels weird. To me weirder than sending result back via the argument. But if this is the overall opinion I will agree etc.

stellarhoof:
Ya let's just get more eyez on this

src/exampleTypes/TagsQuery/ActionsMenu.js
```diff
@@ -1,57 +1,57 @@
...
+    <Flex
+      style={{ minWidth: 240, padding: 10 }}
+      className="tags-query-actions-menu"
+      column
+      justifyContent="stretch"
+      alignItems="stretch"
+    >
+      {!!_.get('tags.length', node) && (
+        <>
+          <Button onClick={() => wrapper('copy', () => copyTags(node))}>
```
daedalus28: I think we can simplify this if we instead just take a wrapper and assume that the return values are success messages and errors are the error messages. On the user side, we'd have one of these that just logs that with toastify:

https://runkit.com/daedalus28/toasty-futil-aspect

Akrion: Changes done.

src/stories/imdb/stories/searchLayout.js
```diff
@@ -258,6 +260,7 @@ let GreyVestSearchBarStory = theme => (
               path={['root', 'bar']}
               resultsPath={['root', 'results']}
               autoFocus
+              actionWrapper={aspectWrapper}
```
Akrion: Let this here as a test/demo of the idea of `aspectWrapper`

src/utils/futil.js
```diff
@@ -38,3 +38,8 @@ export let expandObject = _.curry((transform, obj) => ({
 export let expandObjectBy = _.curry((key, fn, obj) =>
   expandObject(getWith(fn, key))(obj)
 )
+
+export let aspectWrapper = F.aspect({
```
Akrion: Kept this here for demo purposes.

Akrion:
Nothing is set in stone for this and I am open for suggestions. If we are saying that we would not need any feedback from the `action button` and we will handle all that is needed at `contexture-react` level by disabling buttons etc I am fine with that too. We just need to choose a direction/pattern and go with it.

src/utils/futil.js
```diff
@@ -38,3 +38,9 @@ export let expandObject = _.curry((transform, obj) => ({
 export let expandObjectBy = _.curry((key, fn, obj) =>
   expandObject(getWith(fn, key))(obj)
 )
+
+export let aspectWrapper = open => F.aspect({
```
daedalus28: I think we can simplify this even further if we don't make the wrapper aware of `open` - I don't think end users are going to want to control the open-ness from the aspect

daedalus28: but testing this idea here was really smart 👏

Akrion: Changes done.

stellarhoof:
@Akrion You can do `_.has('clipboard', navigator)` or something and disable the button accordingly. Haven't tested in Safari since I don't have a Mac. The second action is a `tree.mutate`. To my earlier point on the blurry lines of error handling: we do contexture stuff like this all over the place so I see no point in wrapping this one.

Akrion:
@stellarhoof I do see your point however I also do believe the user should get a feedback saying that his tags were (or were not) copied into the clipboard. It is a common pattern and you kind of intuitively expect it when you do `copy` to clipboard. Otherwise we really are leaving the user guessing did it work or not and having him "testing" if it did via pasting somewhere. 

So I am not aware of many other places where such a feedback is necessary but if there are we should discuss them as well. It might be that we should also wrap some of them as well. 

Now I am not saying this "wrapper" approach is the best but it seems to be the most out of the box we could do since we already have the AOP support in `Futil` with the `aspect`. Maybe we have a better way to wrap or expose the action result etc ...

stellarhoof:
So do you want to 1) Always have some visual feedback on whether copying something was successful or 2) Show a toast when there's an error? If we want 1) we'd be better off with passing a custom copy function everywhere that displays a toast on success. Another advantage of this is that we can use a polyfill to make sure we always succeed.

src/exampleTypes/TagsQuery/ActionsMenu.js
```diff
@@ -7,7 +7,13 @@ import { withTheme } from '../../utils/theme'
 import { Flex } from '../../greyVest'
 import { copyTags, tagTerm } from './utils'
 
-let ActionsMenu = ({ node, tree, open, theme: { Button, Checkbox } }) => (
+let ActionsMenu = ({
+  node,
+  tree,
+  open,
+  theme: { Button, Checkbox },
+  actionWrapper = _.noop,
```
daedalus28: I think should default to `_.identity`. If you omit `actionWrapper`, `_.noop` returns undefined which would prevent the onClicks from doing anything.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "contexture-react",
-  "version": "2.7.2",
+  "version": "2.7.3",
```
daedalus28: sorry for the churn, just need to update the version

daedalus28:
@stellarhoof the point of this PR is to give visible feedback for these buttons, so I think this is the right approach. We're actually less concerned about errors on copy, and more concerned that the success case currently gives no visual feedback so users have no idea if it worked.

stellarhoof:
@daedalus28 See my previous comment https://github.com/smartprocure/contexture-react/pull/354#issuecomment-552020185

Akrion:
Fixed

PR #356, opened 2019-11-05
Debounce popover onClose

michsa:
This PR fixes an issue where popovers couldn't be toggled off from outside buttons, because the popover's OutsideClickHandler would immediately fire and close it, and then the toggle would simply open it again.

Note that as a side effect of this change, _all_ buttons that open the popover will now "toggle", since the OutsideClickHandler event will still fire first and the debounce will prevent the button's click handler from reopening the popover.

Ideally, we'd want popovers to be able to handle clicks on their trigger component _before_ the internal OutsideClickHandler, to enable both of these use cases: a button that toggles the popover, or a button that can only open it. This would require the popover to be aware of its trigger component, however, which would in turn require an API change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 2.6.1
+* GreyVest: debounce Popover's `onClose` handler (fixes a bug where toggling a popover's open state just kept the popver open)
+
 # 2.6.0
 * FilterList: remove refresh icon & replace it with an "apply filter" button
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "contexture-react",
-  "version": "2.6.0",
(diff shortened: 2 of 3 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 224d040c163f5a4fd149283ea2dd65f6c6a684be
</p>

PR #366, opened 2019-11-15
Add support for the buttonAttributes to the SearchBar

### LINKED ISSUE #365: TagsQuerySearchBar to support passing of buttonAttributes
For some of the tours there is a need to deferenciate between the spending search button and the bids one. Currently `gv-button` is used as a target but it seems that is not helpful enough since that is also the class for half the buttons on the page. 

This ticket proposes the addition of another property to the `TagsQuerySearchBar` which will allow to decorate the `search button` with attributes ... e.g.

`buttonAttributes={{["data-attribute"]: "attribute1"}}`

The final usage would be:

```
<TagsQuerySearchBar
    tree={tree}
    path={['root', 'bar']}
    resultsPath={['root', 'results']}
    autoFocus
    actionWrapper={aspectWrapper}
    buttonAttributes={{["data-tour"]: "spendingButton"}}
 />
```

Akrion:
Fixes #365

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 # 2.9.2
-* Add support for the "buttonAtrribute" prop to the SearchBar
+* Add support for the "searchButtonProps" prop to the SearchBar
 
 # 2.9.1
 * Update `contexture-client` dev dependency and stories to use newly observable `disableAutoUpdate`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# 2.9.2
+* Add support for the "buttonAtrribute" prop to the SearchBar
+
 # 2.9.1
 * Update `contexture-client` dev dependency and stories to use newly observable `disableAutoUpdate`
 * Add a GreyVest Modal story to demonstrate opening a modal from inside a popover
--- a/src/exampleTypes/TagsQuerySearchBar.js
+++ b/src/exampleTypes/TagsQuerySearchBar.js
@@ -44,12 +44,12 @@ let AnimatedButton = ({ disabled, style, className, ...props }) => (
   />
 )
 
-let SearchButton = observer(({ tree, resultsPath, buttonAttributes }) => (
+let SearchButton = observer(({ tree, resultsPath, searchButtonProps }) => (
   <AnimatedButton
     disabled={!tree.getNode(resultsPath).markedForUpdate}
     onClick={tree.triggerUpdate}
     style={buttonStyle}
-    {...buttonAttributes}
+    {...searchButtonProps}
   >
     Search
   </AnimatedButton>
@@ -60,7 +60,7 @@ let SearchBar = ({
   node,
   resultsPath,
   actionWrapper,
-  buttonProps,
+  searchButtonProps,
 }) => {
   let collapse = React.useState(true)
   let popoverState = useLensObject({ open: false, tagOpen: '' })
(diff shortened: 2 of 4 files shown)


src/exampleTypes/TagsQuerySearchBar.js
```diff
@@ -82,7 +89,11 @@ let SearchBar = ({ tree, node, resultsPath, actionWrapper }) => {
           />
         </Box>
         {tree.disableAutoUpdate && (
-          <SearchButton tree={tree} resultsPath={resultsPath} />
+          <SearchButton
```
stellarhoof: This is a very specific use case. Can't we parametrize `SearchButton` itself?

Akrion: @stellarhoof can you elaborate more how that is better and different that this? Also what is the current issue you see if we pass another property to the `TagsQuerySearchBar`. Not sure I follow how the very specific use case is a problem here. It is just another prop passed through to eventually the `SearchButton`.

stellarhoof: Here's how I usually look at this: you've got a component with internal components and you want to forward some props to the internal ones. The line between doing that or parametrizing the component wholesale is kinda blurry. Keeping everything else constant, if you look at the component API and the prop feels kinda arbitrary then you might be better off passing the component itself.

If I see `<TagsQuerySearchBar buttonAttributes={buttonAttributes}/>` my first reaction would be 1) Oh that's kinda specific and 2) Which button gets this? 3) Why not just button props? etc... whereas this `<TagsQuerySearchBar SearchButton={<SearchButton ... />} />` is self-evident. Of course, the second usage would be annoying if we had to use it like that everywhere, but given that this is a very specific use case I think the second approach is better.

Akrion: @stellarhoof we would have to use it everywhere :). In the given request they want to add attributes to the search button for all pages. So spending/bids/contacts would have to have the SearchButton passed ... and we practically do not use anywhere else the `TagsQuerySearchBar`. I agree with your reasoning just not sure if in this case it makes sense since instead of passing an attribute I have to now pass the entire button. Not to mention that the actual `SearchButton` is not even exported from `contexture-react` and we would have to do that as well.

stellarhoof: True, and I'd have no feedback if this was on the bid-search side, but contexture-react is a bit more generic, so having a prop like this for a specific use case on our main product is kinda yucky

Akrion: OK @michsa do you agree as well on @stellarhoof suggestion? If you guys both are on the same opinion I would obviously listen and change it :)

Akrion: Changes made as per the discussion

src/exampleTypes/TagsQuerySearchBar.js
```diff
@@ -60,7 +60,7 @@ let SearchBar = ({
   node,
   resultsPath,
   actionWrapper,
-  buttonAttributes,
+  buttonProps,
```
stellarhoof: `searchButtonProps`?

Akrion: You got it.

src/stories/imdb/stories/searchLayout.js
```diff
@@ -259,6 +259,7 @@ let GreyVestSearchBarStory = theme => (
               resultsPath={['root', 'results']}
               autoFocus
               actionWrapper={aspectWrapper}
+              searchButtonProps={{ ['data-attribute']: 'attribute1' }}
```
Akrion: Just as a "test"

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+# 2.9.2
+* Add support for the "buttonAtrribute" prop to the SearchBar
```
stellarhoof: Also update this

PR #380, opened 2020-02-04
futil import fixes

ltchris:
From the 2.7.1 update [here](https://github.com/smartprocure/contexture-react/blob/ddf47ad6fc476fcea05bb77dffc7d8f3b197b12f/CHANGELOG.mdx#271) and [the commits](https://github.com/smartprocure/contexture-react/commit/c8c8b2d9464bc09697e64e9dcb0a809f090111db), it looks like these were missed when making the transition from `futil-js` to `futil`.

### DIFF
--- a/src/exampleTypes/FacetSelect.js
+++ b/src/exampleTypes/FacetSelect.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import F from 'futil-js'
+import F from 'futil'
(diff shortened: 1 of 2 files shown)


daedalus28:
Thanks for the contribution! I merged this into the memory table branch that’s about to get merged so you don’t need to do a version bump

PR #384, opened 2020-02-20
Load more results

michsa:
ResultTableFooter (and by extension ResultTable) now shows a 'Load More' button in its pagination control if the results node holds a `skipCount` flag. It also keeps track of the running total on its own, so it can paginate as normal through any already-discovered records. 

Right now this is only relevant for contexture-mongo, since it's the only provider whose results node supports `skipCount`.

This PR also adds TableFooter and PageSize from grey-vest 3.0, and updates ResultTableFooter to use them (otherwise any changes made here would have to be rewritten in a couple of weeks). Some of the code for 'load more' lives in TableFooter and will have to be ported back to grey-vest 3.0, but PageSize was copied unchanged.

### DIFF
--- a/src/greyVest/TableFooter.js
+++ b/src/greyVest/TableFooter.js
@@ -6,24 +6,28 @@ import PagerItem from './PagerItem'
 import PageSize from './PageSize'
 import Flex from './Flex'
 
-let getRange = (first, last) => (first >= last ? last : `${first}-${last}`)
+let PageDetails = ({ startRecord, endRecord, totalRecords, hasMore }) => (
+  <span style={{ flex: '0 1 30%', textAlign: 'right' }}>
+    <b>Showing </b>
+    {startRecord >= endRecord ? endRecord : `${startRecord}-${endRecord}`}
+    {F.isNotNil(totalRecords) && ` of ${totalRecords}${hasMore ? '+' : ''}`}
+  </span>
+)
 
-let getTotal = (totalRecords, hasMore) =>
-  F.isNotNil(totalRecords) ? `of ${totalRecords}${hasMore ? '+' : ''}` : ''
-
+// Accepts either `totalRecords` or 
 let TableFooter = ({
   page = 1,
   onChangePage,
   pageSize,
   onChangePageSize,
   pageSizeOptions,
-  startRecord = pageSize * (page - 1) + 1,
-  endRecord = 0,
   hasMore,
-  totalRecords = hasMore ? undefined : endRecord,
+  totalRecords,
+  startRecord = pageSize * (page - 1) + 1,
+  endRecord = totalRecords ? _.min([totalRecords, pageSize * page]) : 0,
   ...props
 }) => {
-  if (totalRecords) endRecord = _.min([totalRecords, endRecord])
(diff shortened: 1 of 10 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 31811a8b51bf37ede7e7e4aa323d51f0aeae442b
</p>

src/greyVest/TableFooter.js
```diff
@@ -0,0 +1,57 @@
...
+            style={{ margin: '0 8px', paddingLeft: 12, paddingRight: 12 }}
+            onClick={() => onChangePage(page + 1)}
+          >
+            Load More...
+          </PagerItem>
+        )}
+      </Flex>
+      <span style={{ flex: 1, textAlign: 'right' }}>
+        <b>Showing</b> {showing(totalRecords, page, pageSize)}
+        {!hasMore && ` of ${totalRecords}`}
```
daedalus28: It occurs to me that this isn't obvious - it's not intuitive that when skipCount is false, hasMore will always be falsey. We should probably change that and make this check based on `node.skipCount`

src/greyVest/TableFooter.js
```diff
@@ -0,0 +1,57 @@
...
+import _ from 'lodash/fp'
+import Pager from './Pager'
+import PagerItem from './PagerItem'
+import PageSize from './PageSize'
+import Flex from './Flex'
+
+let showing = (totalRecords, page, pageSize) => {
+  let min = _.min([totalRecords, pageSize * (page - 1) + 1])
+  let max = _.min([totalRecords, pageSize * page])
+  return min >= max ? max : `${min}-${max}`
```
daedalus28: This check is confusing - `min > max` feels like a bug even if it makes logical sense

src/greyVest/TableFooter.js
michsa: This is a GreyVest component - a fairly specialized one, but it still shouldn't know more about contexture than it has to.

From the perspective of TableFooter, `hasMore` is just a flag that indicates we shouldn't show ` of {totalRecords}` because `totalRecords` may be incorrect (and TableFooter doesn't care about why it's falsy - it's a flag, so falsy is the default).

daedalus28: Understood, but the logic ends up confusing - this would need to be passed a falsely value hasMore even when there is more data when the count is provided. Perhaps another alternative would be to call the prop something like hideCount on the pager which is passed from contexture react based on skipCount

PR #385, opened 2020-02-20
Fix greyVest import documentation

ltchris:
The theme story example has an import for `greyVestTheme` instead of `greyVest`, which might be intentional since it is just an example, but it could be a confusing to new users as to what the import is for the grey vest theme.


daedalus28:
Thanks for the contribution! I merged this into the memory table branch that’s about to get merged so you don’t need to do a version bump

PR #401, opened 2020-05-01
Add bool/exists display prop for customizations

Akrion:
- Adds `display`support to the bool and exists types for customizing the labels
- Adds support for the `either` option (which is value `null` and returns all records)
- Adds stories for both types customization options

Related to https://github.com/smartprocure/spark/issues/5469

### DIFF
--- a/src/exampleTypes/Bool.js
+++ b/src/exampleTypes/Bool.js
@@ -1,36 +1,4 @@
 import React from 'react'
-import _ from 'lodash/fp'
-import { contexturify } from '../utils/hoc'
+import BooleanType from '../purgatory/BooleanType'
 
-let getValue = (value, either) => _.isBoolean(value)
-  ? value
-  : either ? null : false
-
-let Bool = ({
-  tree,
-  node,
-  display = () => ['Yes', 'No', 'Either'],
-  theme: { RadioList }
-}) => {
-  let [yes, no, either] = display()
-  return (
-    <div className="contexture-bool">
-      <RadioList
-        value={getValue(node.value, either)}
-        onChange={value => {
-          tree.mutate(node.path, {
-            value: getValue(value, either)
-          })
-        }}
-        options={[
-          { label: yes, value: true }, 
-          { label: no, value: false }, 
-          ...(either ? [{ label: either, value: null}] : [])
-        ]}
-      />
-    </div>
-
-  )
-}
-
-export default contexturify(Bool)
+export default ({...props }) => <BooleanType {...props} />
--- a/src/exampleTypes/Bool.js
+++ b/src/exampleTypes/Bool.js
@@ -1,5 +1,4 @@
 import React from 'react'
-import F from 'futil'
 import _ from 'lodash/fp'
 import { contexturify } from '../utils/hoc'
 
@@ -30,6 +29,7 @@ let Bool = ({
         ]}
       />
     </div>
+
   )
 }
 
--- a/src/exampleTypes/Bool.js
+++ b/src/exampleTypes/Bool.js
@@ -1,22 +1,33 @@
 import React from 'react'
 import F from 'futil'
+import _ from 'lodash/fp'
 import { contexturify } from '../utils/hoc'
+
+let getValue = (value, either) => _.isBoolean(value)
+  ? value
+  : either ? null : false
 
 let Bool = ({
   tree,
   node,
-  display = (options=['yes', 'no']) => options,
+  display = () => ['Yes', 'No', 'Either'],
   theme: { RadioList }
 }) => {
-  let options = display()
+  let [yes, no, either] = display()
   return (
     <div className="contexture-bool">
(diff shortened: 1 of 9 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


Take a look at this commit to see what happened in detail: fb329ba8116226ab7dada2ec00c4b809b1880ffa


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against f61d4a9c20ec4228529c16493d2c36338ab4715f
</p>

src/exampleTypes/Exists.js
```diff
@@ -2,14 +2,14 @@ import React from 'react'
 import F from 'futil'
 import { contexturify } from '../utils/hoc'
 
-let Exists = ({ tree, node, theme: { RadioList } }) => (
+let Exists = ({ tree, node, options=['exists', 'doesNotExist'], theme: { RadioList } }) => (
```
Akrion: This is basically the same component as bool. Wonder if it makes sense to extract the common stuff to one component and use it in both ... I did not do it since I can not imagine that was not discussed before ect.

src/exampleTypes/Bool.js
```diff
@@ -1,17 +1,34 @@
...
-    />
-  </div>
-)
+let getValue = (value, either) =>
+  _.isBoolean(value) ? value : either ? null : false
+
+let Bool = ({
+  tree,
+  node,
+  display = () => ['Yes', 'No', 'Either'],
```
stellarhoof, line 11: Shouldn't this be yes/no/none/both?

src/exampleTypes/Bool.js
```diff
@@ -1,17 +1,34 @@
...
-      value={node.value ? 'yes' : 'no'}
-      onChange={value => {
-        tree.mutate(node.path, { value: value === 'yes' })
-      }}
-      options={F.autoLabelOptions(['yes', 'no'])}
-    />
-  </div>
-)
+let getValue = (value, either) =>
+  _.isBoolean(value) ? value : either ? null : false
```
daedalus28, line 6: This might be simpler with a single check instead of a nested ternary:

```js
let getValue = (value, either) => (_.isNil(value) && either)  ? null : !!value
```

The main implementation difference is that this new version will continue to treat truthy values as `true`, whereas the one in the PR here would treat something like `5` as `false`

src/exampleTypes/Bool.js
daedalus28, line 11: this API is different from `facet` - let's have them match and have display be a function that takes the value and returns label

```js
value => _.isNil(value) ? 'Either' : value ? 'Yes' : 'No'
```

You're welcome to export a utility that takes an array:

```js
let generateBoolDisplay: ([yes, no, either]) => value => _.isNil(value) ? either : value ? yes : no
```

But the actual API should follow the convention we have in `facet` and results table of display being a function that takes a value and returns a label. With that said, the issue with relying on array order is that there is no obvious ordering - you'd have to look it up in the docs either way, where as a `value ->label` function follows the convention and doesn't require you to remember an arbitrary order (even if it's more verbose)

src/exampleTypes/Bool.js
daedalus28: Also, it might make sense to curry and flip this, so `either` comes first, but that's up to you since this is an internal util

src/exampleTypes/Exists.js
```diff
@@ -1,17 +1,34 @@
...
-        tree.mutate(node.path, { value: value === 'exists' })
-      }}
-      options={F.autoLabelOptions(['exists', 'doesNotExist'])}
-    />
-  </div>
-)
+let getValue = (value, either) =>
+  _.isBoolean(value) ? value : either ? null : false
+
+let Exists = ({
```
daedalus28, line 8: yeah, you're right - this is line for line identitical with bool, except the className and default display. We should definitely factor this out

Akrion:
Changes made

src/exampleTypes/Bool.js
```diff
@@ -1,17 +1,34 @@
...
+let Bool = ({
+  tree,
+  node,
+  display = () => ['Yes', 'No', 'Either'],
+  theme: { RadioList },
+}) => {
+  let [yes, no, either] = display()
+  return (
+    <div className="contexture-bool">
+      <RadioList
```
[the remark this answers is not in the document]
daedalus28: they are all mutually exclusive options, so a radio list makes sense

PR #426, opened 2020-08-05
Rename Tooltip name from toggle filter to show/hide filter 

chris110408:
fix the issue https://github.com/smartprocure/spark/issues/6920


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Branch being merged does not follow [Git Flow](http://nvie.com/posts/a-successful-git-branching-model/)
  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against ab42863660a1d7fcfcaf4f488704a1d4c1267368
</p>

src/purgatory/ToggleFiltersButton.js
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import { withTheme } from '../utils/theme'
 
 let ToggleFiltersButton = ({ onClick, theme: { AlternateButton, Icon } }) => (
-  <AlternateButton title="Toggle Filters" onClick={onClick}>
+  <AlternateButton title="Show Filters" onClick={onClick}>
```
daedalus28, line 5: A generic `ToggleFiltersButton` shouldn't assume that it's always `Show`. Even that happens to be true in our use cases, that's not the assumption you'd make for this. Ideally, we'd make it dynamic even though this currently doesn't know about the current value of show/hide.

daedalus28: After talking through this, the implementation would get really complicated because mode is actually a 3 way toggle. Let just call this `ShowFiltersButton` instead so it's super clear. This is, in theory, a breaking change - but real end users would just use `ToggleFiltersHeader` so this component never really needed to be exported anyway. We'll pretend like it never was 😄

PR #436, opened 2020-08-25
Feature/430 - update signature of loadHereOptions to support passing country as second arg, backwards compatible

### LINKED ISSUE #430: Support passing country to loadHereOptions
Right now `loadHereOptions` comes out of the box with a default config that biases the results to the US. You can change this, but you have to pass in a whole separate config object, which is onerous if you're relying on the default config and have no idea what the rest of its values are.

Need to change the signature from inputValue and an optional hereConfig to inputValue, optional country, optional hereConfig. 
* The API should allow getting unbiased results if one wants them
* in order to avoid a breaking change one should get results biased to the US if one only passes `inputValue`
* in order to avoid a breaking change the case where `inputValue, hereConfig` is passed should be handled as though `inputValue, hereConfig.country, hereConfig` had been passed
* so, to be consistent with the above constraints the API should be that one must explicitly pass `null` as the second argument if one wants unbiased results

doug-patterson:
Fixes #430 

### Description
* lets you re-bias HereMaps address completion options to another country.
* behavior unchanged for existing code that passes only 1 or 2 arguments

### DIFF
--- a/CHANGELOG.mdx
+++ b/CHANGELOG.mdx
@@ -1,3 +1,6 @@
+### 2.29.0
+* Geo: support passing country as second argument to `loadHereOptions`
+
 ### 2.28.4
 * GreyVest: add className `gv-table-parent` to table parent `div`
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "contexture-react",
-  "version": "2.28.4",
+  "version": "2.29.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "contexture-react",
-  "version": "2.28.4",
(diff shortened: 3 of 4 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 490fb2c50fcad3b9b069eda5154564250a8cc6c8
</p>

doug-patterson:
@daedalus28 ready for review

doug-patterson:
ready for review

PR #438, opened 2020-08-29
Feature/473 filter list animation

### LINKED ISSUE #437: Add collapsing animation for `FilterList` component
When the user opens and closes sections in Contexture filters there should be a subtle animation for collapsing and expanding a section

rma4ok:
1. Making Filter List sections slide 🏄‍♂️🏄‍♂️🏄‍♂️
1. Slowly dropping shadows between sections ➫➫➫
1. Compensating for `.gv-text-button` extra padding with negative margins, to keep within line-height so it doesn't jump

Closes #437


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


Take a look at this commit to see what happened in detail: 05db21fc22a011e54fc718d1e01ee5943ea16515


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against f9ff5f1844b1e6fb3e59721f1cf9fbe0fd2a4296
</p>

src/greyVest/Table.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 
 let Table = x => (
-  <div clasName="gv-table-parent" style={{ overflow: 'auto' }}>
+  <div className="gv-table-parent" style={{ overflow: 'auto' }}>
```
daedalus28, line 4: nice catch!

src/FilterList.js
```diff
@@ -173,11 +173,33 @@ let FilterList = _.flow(
...
+              className="filter-list-item"
+              style={{
+                willChange: 'box-shadow',
+                transition: `box-shadow .4s ease-${
+                  child.paused ? 'in' : 'out'
+                }`,
+                boxShadow:
+                  child !== _.last(_.get('children', node)) && !child.paused
+                    ? 'rgba(0, 0, 0, 0.1) 0 -16px 20px -20px inset'
+                    : 'rgba(0, 0, 0, 0.1) 0 0 20px -20px inset',
```
daedalus28, line 187: This seems like the realm of grey-vest, not contexture-react (which is still packaged in this lib). Alternate themes might not want an animated box shadow. The first thing that comes to mind would be to animate on and off a class that the grey-vest style sheet can target, but we did intend to move to emotion + theme UI in grey vest and kill the css in v3. I'm open to suggestions.

src/FilterList.js
```diff
@@ -173,11 +173,33 @@ let FilterList = _.flow(
...
+              <div
+                style={{
+                  maxHeight: child.paused ? 0 : '80vh',
+                  overflowY: child.paused ? 'hidden' : 'auto',
+                  willChange: 'max-height',
+                  transition: `max-height .3s ease-${
+                    child.paused ? 'out' : 'in'
+                  }`,
+                }}
+              >
```
daedalus28, line 202: same as above - perhaps we can factor out an accordion component that can go on the theme, unified with an accordion we add to grey vest

src/greyVest/Style.js
```diff
@@ -256,6 +256,7 @@ export default () => (
       .gv-text-button {
         border-radius: 100px;
         padding: 5px;
+        margin: -4px 0 -6px;
```
daedalus28, line 259: We should document why we're doing this. It's a reasonable solution, but not the reasoning isn't obvious from this line

daedalus28:
Also, don't forget to bump the package again. The prior merged PR already took us to 2.29

src/greyVest/Style.js
stellarhoof: This feel wrong. Why is this button specifying its margins?

rma4ok: Added comment

src/FilterList.js
rma4ok: Moved to GreyVest theme

src/FilterList.js
rma4ok: Moved to GreyVest theme

src/greyVest/Style.js
rma4ok: @stellarhoof this meant to be inline text/icon button. It specifies extra padding to have a larger hover area. But if padding is added to the vertical height this button breaks out of vertical line-height. Here is this button on hover, so you can see how much vertical space it takes with that padding:
Screen Shot 2020-08-31 at 1 16 26 PM

Without this negative margin, it makes the whole section header jump in height when you open the section

stellarhoof: In that case we can set a min height on that filter header or toggle the button's visibility or something else. I don't think top-level dom nodes should specify their margins since that's a parent layout concern. If we make this change, every text button will define its outer spacing.

rma4ok: Alright, that makes sense. I felt like fixing it right next to the cause and make text buttons fit into any text line without bloating it. But your point about concern separation is also valid. I will remove this and change the button's visibility and inner padding in the filter header.

rma4ok:
@daedalus28 I've moved Accordion into own Grey Vest component 
@stellarhoof removed shadow separation for open sections

stellarhoof:
@rma4ok Got conflicts :)

src/greyVest/Expandable.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import Flex from './Flex'
+import Icon from './Icon'
+
+let Expandable = ({ isOpen, className, style, Label, children, onClick }) => (
+  <div
+    className={`gv-accordion ${isOpen ? 'expanded' : ''} ${className}`}
```
daedalus28, line 7: we should replace all references to `gv-accordion` with `gv-expandable`

src/greyVest/Style.js
```diff
@@ -30,6 +30,33 @@ export default () => (
         margin: 30px 0;
       }
 
+      /* Accordion */
+      .gv-accordion-icon {
```
daedalus28, line 34: accordion -> expandable

src/themes/greyVest/Style.js
```diff
@@ -109,6 +110,19 @@ export default () => (
...
+      .filter-field-label-icon {
+        opacity: 0;
+        will-change: opacity;
+        transition: opacity .2s ease-in;
+      }
+      .filter-list-item.expanded .filter-field-label-icon {
+        opacity: 1;
+        transition: opacity .2s ease-out;
+      }
+      .filter-list-item .gv-accordion-icon {
```
daedalus28, line 124: accordion -> expanded

src/FilterList.js
```diff
@@ -173,41 +170,50 @@ let FilterList = _.flow(
...
+              className="filter-list-item"
+              isOpen={!child.paused}
+              Label={
+                <Label tree={tree} node={child} fields={fields}>
+                  {mapNodeToLabel(child, fields)}
+                </Label>
+              }
+              onClick={() =>
+                tree &&
+                node &&
```
daedalus28, line 184: this isn't the right check - the prior implementation checked tree and node because it was toggling node.paused at node.path. This now uses child, so it should check for child instead, right?

src/FilterList.js
```diff
@@ -173,41 +170,50 @@ let FilterList = _.flow(
...
+                tree.mutate(child.path, { paused: !child.paused })
+              }
+            >
+              <div className="filter-list-item-contents">
+                <Dynamic
```
daedalus28, line 189: It occurs to me that this might have unexpected consequences since we will be rendering node all the time even if they're paused. It's _probably_ not an issue, but we'll need to test this carefully.

src/FilterList.js
rma4ok: Yeah that's right

rma4ok: I am skipping the child check tho, it's always present in given context

PR #441, opened 2020-09-23
Feature/popover component

rma4ok:
* Using `reactjs-popup` as the basis for Popover component
* Fixed positioning of the Popover and it being cut off in Results Table Header
* Integrating `open` lense and `openBinding` for controlling Popover state
* Refactored components that use Popover to leverage state management to the Popover itself
* Using arrows to indicate an element showing the Popover
* Also: Changes section indicator animation for Expandable component

### DIFF
--- a/src/exampleTypes/ExpandableTagsQuery/index.js
+++ b/src/exampleTypes/ExpandableTagsQuery/index.js
@@ -11,8 +11,8 @@ let collapsedStyle = {
   overflowY: 'auto',
 }
 
-// eslint-disable-next-line no-unused-vars
 let ExpandableTagsQuery = ({
+  // eslint-disable-next-line no-unused-vars
   measure,
   measureRef,
   contentRect,
--- a/src/exampleTypes/ExpandableTagsQuery/index.js
+++ b/src/exampleTypes/ExpandableTagsQuery/index.js
@@ -11,7 +11,8 @@ let collapsedStyle = {
   overflowY: 'auto',
 }
 
-let ExpandableTagsQuery = ({ measureRef, contentRect, collapse, ...props }) => (
+// eslint-disable-next-line no-unused-vars
+let ExpandableTagsQuery = ({ measure, measureRef, contentRect, collapse, ...props }) => (
   <>
     <div style={F.view(collapse) ? collapsedStyle : {}}>
       <div ref={measureRef}>
--- a/src/exampleTypes/ExpandableTagsQuery/index.js
+++ b/src/exampleTypes/ExpandableTagsQuery/index.js
@@ -11,18 +11,11 @@ let collapsedStyle = {
   overflowY: 'auto',
 }
 
-let ExpandableTagsQuery = ({
-  // eslint-disable-next-line no-unused-vars
-  measure,
-  measureRef,
-  contentRect,
-  collapse,
-  ...props
-}) => (
+let ExpandableTagsQuery = ({ measureRef, contentRect, collapse, ...props }) => (
   <>
     <div style={F.view(collapse) ? collapsedStyle : {}}>
       <div ref={measureRef}>
-        <TagsQuery {...props} />
+        <TagsQuery {..._.omit('measure', props)} />
       </div>
     </div>
     {F.view(collapse) && contentRect.entry.height > innerHeight && (
--- a/src/exampleTypes/ExpandableTagsQuery/index.js
+++ b/src/exampleTypes/ExpandableTagsQuery/index.js
@@ -12,7 +12,13 @@ let collapsedStyle = {
 }
 
 // eslint-disable-next-line no-unused-vars
-let ExpandableTagsQuery = ({ measure, measureRef, contentRect, collapse, ...props }) => (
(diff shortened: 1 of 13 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
      <td>

<details>
  <summary>Header.js</summary>
  <p><a href="https://github.com/smartprocure/contexture-react/blob/feature/popover-component/src/exampleTypes/ResultTable/Header.js" target="_blank">src/exampleTypes/ResultTable/Header.js</a></p>
  
  <strong>'popover' is assigned a value but never used.</strong>
  <code>Line 65: let popover = React.useState(false)</code>
</details>
  </td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against bb65ec5199ebb7592c711f380ca994d5072fd7fa
</p>

src/greyVest/Popover.js
```diff
@@ -1,32 +1,44 @@
...
-          }}
-        >
-          {children}
-        </div>
-      </div>
-    </OutsideClickHandler>
-  )
+let Popover = ({
+  // eslint-disable-next-line no-unused-vars
+  open, // lens
```
daedalus28, line 10: if you're using expandProp with `openBinding`, the idea is that you should be coding against `isOpen`/`onClose`. `openBinding` is about adding lens support to a standard API - you should pick one as the canonical API you code against and then at most adapt to the other. In general, we've tried to move away from lenses a bit as most of the team find the abstraction to be more overhead than it's worth which is actually where that pattern of expanding with a binding came from - we add lens support on top, but most of the code doesn't deal with them

src/exampleTypes/ExpandableTagsQuery/index.js
```diff
@@ -11,7 +11,14 @@ let collapsedStyle = {
   overflowY: 'auto',
 }
 
-let ExpandableTagsQuery = ({ measureRef, contentRect, collapse, ...props }) => (
+let ExpandableTagsQuery = ({
+  // eslint-disable-next-line no-unused-vars
+  measure,
```
daedalus28, line 16: Why destructure measure if it's not used? Is it so it's omitted on `props` before passing to `TagsQuery`?

daedalus28 (changes requested):

src/greyVest/Popover.js
```diff
@@ -1,32 +1,44 @@
...
+  style,
+  trigger,
+  arrow,
+  position,
+  children,
+  closeOnPopoverClick = true,
+  ...props
+}) => (
+  <Popup
+    open={open && isOpen} // external state if lens is passed
```
stellarhoof, line 21: `open` is expanded into `isOpen` and `onClose`: https://github.com/smartprocure/contexture-react/blob/f9ceefe1cd1ad5a8c16e245043a460ed8e654590/src/greyVest/utils.js#L3 so we can just use `isOpen` here.

Also, according to the docs if you use `open` you need to pass `onClose` as well (https://react-popup.elazizi.com/controlled-popup).

Here's the API of our current popover:

1. `<Popover open={lens} />`
2. `<Popover isOpen={bool} onClose={fn} />`

We need to support that + `trigger` so

- If `isOpen` is not undefined, pass `open={isOpen}` and `onClose={onClose}` to `Popup`
- if `trigger` is not undefined, don't pass `open` or `onClose`

src/exampleTypes/ExpandableTagsQuery/index.js
rma4ok: Yeah, otherwise it gets passed to an `input` inside of `TagsQuery` and shows React error

src/greyVest/Popover.js
rma4ok: `onClose` is getting passed to the Popup in the `...props`

src/exampleTypes/ExpandableTagsQuery/index.js
rma4ok: I've used `_.omit` instead

src/greyVest/Popover.js
rma4ok: Changed to using `onClose` explicitly and added comments

src/greyVest/Popover.js
rma4ok: Changed to using `explodeProp` util I just created. It omits the `open` key

daedalus28 (approved):

PR #477, opened 2021-03-12
[feature] MemoryTable: updating memory service with fresh data

rma4ok:
* [x] MemoryTable: updating memory service with fresh data on render
* [x] ResultTable: hiding footer with `hideFooter` prop ~~and applying `cellProps` and `headerCellProps` from field schema~~

### DIFF
--- a/src/MemoryTable.js
+++ b/src/MemoryTable.js
@@ -1,8 +1,6 @@
-import _ from 'lodash/fp'
 import Contexture from 'contexture'
 import memory from 'contexture/src/provider-memory'
 import types from 'contexture/src/provider-memory/exampleTypes'
-import { observer } from 'mobx-react'
 import React from 'react'
 import ContextureMobx from './utils/contexture-mobx'
 import { componentForType } from './utils/schema'
@@ -10,32 +8,27 @@ import { ResultTable, TypeMap } from './exampleTypes'
 
 export let useMemoryTree = ({
   records,
-  schema = 'data',
-  fields,
   debug,
   resultNode = {
     pageSize: 50,
   },
   criteriaNodes = [],
-  childrenNodes = [],
 } = {}) => {
-  let include = _.map('field', fields)
   let [memoryStorage] = React.useState({ records: [] })
-  let [tree] = React.useState(
+  let [tree] = React.useState( () =>
     ContextureMobx({
       disableAutoUpdate: true,
       service: Contexture({
         debug,
-        schemas: { [schema]: { memory: memoryStorage } },
+        schemas: { data: { memory: memoryStorage } },
         providers: { memory: { ...memory, types: types() } },
       }),
     })({
       key: 'root',
       schema: 'data',
       children: [
-        { key: 'results', type: 'results', include, ...resultNode },
+        { key: 'results', type: 'results', ...resultNode },
         { key: 'criteria', type: 'group', children: criteriaNodes },
-        ...childrenNodes,
       ],
     })
   )
@@ -57,16 +50,13 @@ let MemoryTable = ({
   debug,
   resultNode,
   criteriaNodes,
-  childrenNodes,
   ...props
 }) => {
   let tree = useMemoryTree({
     records: data,
-    fields,
     debug,
     resultNode,
     criteriaNodes,
-    childrenNodes,
   })
 
   return (
@@ -79,4 +69,4 @@ let MemoryTable = ({
   )
 }
 
-export default observer(MemoryTable)
+export default MemoryTable
--- a/src/MemoryTable.js
+++ b/src/MemoryTable.js
@@ -7,26 +7,30 @@ import ContextureMobx from './utils/contexture-mobx'
 import { componentForType } from './utils/schema'
 import { ResultTable, TypeMap } from './exampleTypes'
 
-export let memoryService = (records, { schema, debug } = {}) => {
-  let storage = { records }
-  return {
-    updateMemory(records) {
-      storage.records = records
-    },
-    service: Contexture({
-      debug,
-      // Hack to effectively set a default schema: if our tree root doesn't have
-      // a `schema` property, it will get the schema at key `undefined`.
-      schemas: { [schema]: { memory: storage } },
-      providers: { memory: { ...memory, types: types() } },
-    }),
+export let memoryService = (records, { schema, debug } = {}) =>
+  Contexture({
+    debug,
+    // Hack to effectively set a default schema: if our tree root doesn't have
+    // a `schema` property, it will get the schema at key `undefined`.
+    schemas: { [schema]: { memory: { records } } },
+    providers: { memory: { ...memory, types: types() } },
+  })
+
+export let useMemory = ({ schema, debug } = {}) => {
+  let [storage] = React.useState({})
+  let [service] = React.useState(Contexture({
+    debug,
+    schemas: { [schema]: { memory: storage } },
+    providers: { memory: { ...memory, types: types() } },
+  }))
+  let updateMemory = records => {
+    storage.records = records
   }
+  return [service, updateMemory]
 }
 
 let MemoryTable = ({ data, fields, debug, include, ...props }) => {
-  let [{ service, updateMemory }] = React.useState(
-    memoryService(data, { schema: 'data', debug })
-  )
+  let [service, updateMemory] = useMemory({ schema: 'data', debug })
   let [tree] = React.useState(
     ContextureMobx({ service })({
       key: 'root',
--- a/src/MemoryTable.js
+++ b/src/MemoryTable.js
@@ -44,7 +44,8 @@ export let useMemoryTree = ({
   // useEffect is not working due to shallow equality check
   if (!_.isEqual(dependency, [resultsNode, criteriaNodes])) {
     setDependency([resultsNode, criteriaNodes])
-    setTree(makeTree())
+    tree = makeTree()
+    setTree(tree)
   }
 
   Promise.resolve(records).then(records => {
--- a/src/MemoryTable.js
+++ b/src/MemoryTable.js
@@ -33,13 +33,14 @@ export let useMemoryTree = ({
     })
   )
 
-  if (records !== memoryStorage.records) {
-    let updateMemory = async records => {
-      memoryStorage.records = await records
+  let updateMemory = async data => {
+    let records = await data
+    if (records !== memoryStorage.records) {
+      memoryStorage.records = records
       tree.refresh(['root'])
     }
-    updateMemory(records)
   }
+  updateMemory(records)
 
   return tree
 }
--- a/src/MemoryTable.js
+++ b/src/MemoryTable.js
@@ -9,7 +9,7 @@ import { componentForType } from './utils/schema'
 import { ResultTable, TypeMap } from './exampleTypes'
 
 export let useMemoryTree = ({
-  records = [],
+  records,
   schema = 'data',
   fields,
   debug,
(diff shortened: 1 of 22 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


Take a look at this commit to see what happened in detail: 123173b46505cd9ce803c1aed34928d178d06bb3


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 923f200069b8bf05ba4d3316f18a107b62fdd0d6
</p>

daedalus28:
It seem like there's some overlap with what you did in contexture-react and the ]`makeMemoryTree` method we have in spark](https://github.com/smartprocure/spark/blob/ac1b48173d864ba887c1633100db9cdda7540645/web/src/util/contexture.js#L85-L125). We should probably unify them and either move it all to the client or contexture-react

rma4ok:
@daedalus28 yeah I think we can make it a part of MemoryTable and use it in spark instead of  makeMemoryTree + ResultTable

src/MemoryTable.js
```diff
@@ -16,8 +16,23 @@ export let memoryService = (records, { schema, debug } = {}) =>
     providers: { memory: { ...memory, types: types() } },
   })
 
+export let useMemory = ({ schema, debug } = {}) => {
+  let [storage] = React.useState({})
+  let [service] = React.useState(
+    Contexture({
```
stellarhoof, line 22: `useState` can take a fn to only compute on first render, which we should do here. Also this looks exactly like the `memoryService` function in this file.

Also further: it looks like `memoryService` is used only in stories and the stories themselves could use `useMemory` instead so I don't think there's a reason for `memoryService` to exist anymore.

stellarhoof (changes requested):

src/MemoryTable.js
rma4ok: Refactored into useMemoryTree

src/MemoryTable.js
```diff
@@ -7,28 +8,67 @@ import ContextureMobx from './utils/contexture-mobx'
...
+        { key: 'criteria', type: 'group', children: criteriaNodes },
+        ...childrenNodes,
       ],
     })
   )
-  tree.refresh(['root'])
+
+  if (records !== memoryStorage.records) {
+    let updateMemory = async records => {
+      memoryStorage.records = await records
```
stellarhoof, line 45: Pretty weird from an API design point of view that we somehow handle a promise here. This should be the caller's responsibility not ours.

src/MemoryTable.js
```diff
@@ -7,28 +8,67 @@ import ContextureMobx from './utils/contexture-mobx'
...
+}) => {
+  let tree = useMemoryTree({
+    records: data,
+    fields,
+    debug,
+    resultNode,
+    criteriaNodes,
+    childrenNodes,
+  })
+
```
stellarhoof, line 71: I think we can make it nicer and more composable with something like this

```javascript
let createResultTableTree = (criteria, results) => ({
  key: 'root',
  schema: 'data',
  children: [
     { key: 'results', type: 'results', include, ...results },
     { key: 'criteria', type: 'group', children: criteria },
  ]
}) 

let MemoryTable = ({ data, fields, criteria, results, ...props }) => {
  let tree = createResultTableTree(criteria, results)
  let client = useMemoryClient({ tree, data, fields })
  return <ResultTable path={['root', 'results']} criteria={['root', 'criteria']} />
}
```

That way users of `useMemoryClient` can pass whatever they want. For example:

```javascript
let client = useMemoryClient({ tree: { type: 'results', pageSize: 100 }, data, fields })
```

src/MemoryTable.js
```diff
@@ -7,28 +8,67 @@ import ContextureMobx from './utils/contexture-mobx'
...
+      memoryStorage.records = await records
+      tree.refresh(['root'])
+    }
+    updateMemory(records)
+  }
+
+  return tree
+}
+
+let MemoryTable = ({
```
stellarhoof, line 54: Can't comment on the code at the end of this file but I'm pretty sure this component doesn't need to be an observer.

src/MemoryTable.js
```diff
@@ -7,28 +8,67 @@ import ContextureMobx from './utils/contexture-mobx'
...
+  schema = 'data',
+  fields,
+  debug,
+  resultNode = {
+    pageSize: 50,
+  },
+  criteriaNodes = [],
+  childrenNodes = [],
+} = {}) => {
+  let include = _.map('field', fields)
```
stellarhoof, line 22: I don't think we need the `include` here if the default is to include everything.

stellarhoof (changes requested):

src/MemoryTable.js
```diff
@@ -7,28 +8,67 @@ import ContextureMobx from './utils/contexture-mobx'
...
+  let [memoryStorage] = React.useState({ records: [] })
   let [tree] = React.useState(
-    ContextureMobx({ service })({
+    ContextureMobx({
```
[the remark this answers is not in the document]
rma4ok: We are using `disableAutoUpdate` here, which disables debounce completely.

src/MemoryTable.js
```diff
@@ -7,28 +8,67 @@ import ContextureMobx from './utils/contexture-mobx'
...
-    // a `schema` property, it will get the schema at key `undefined`.
-    schemas: { [schema]: { memory: { records } } },
-    providers: { memory: { ...memory, types: types() } },
-  })
-
-let MemoryTable = ({ data, fields, debug, include, ...props }) => {
-  let service = memoryService(data, { schema: 'data', debug })
+export let useMemoryTree = ({
+  records,
+  schema = 'data',
```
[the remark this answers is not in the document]
rma4ok: Good call :)

src/MemoryTable.js
```diff
@@ -7,28 +8,67 @@ import ContextureMobx from './utils/contexture-mobx'
...
+    ContextureMobx({
+      disableAutoUpdate: true,
+      service: Contexture({
+        debug,
+        schemas: { [schema]: { memory: memoryStorage } },
+        providers: { memory: { ...memory, types: types() } },
+      }),
+    })({
       key: 'root',
       schema: 'data',
```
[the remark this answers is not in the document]
rma4ok: Now it's hardcoded as `data` in both places

src/exampleTypes/Facet.stories.js
```diff
@@ -19,13 +18,11 @@ export let facetSelect = () => (
...
 
 export let emojiDataset = () => {
-  let data = require('emoji-datasource')
-  let service = memoryService(data)
-  let tree = ContextureMobx({ service })({
-    key: 'root',
-    children: [{ type: 'facet', field: 'category' }, { type: 'results' }],
+  let tree = useMemoryTree({
+    records: require('emoji-datasource'),
+    childrenNodes: [{ type: 'facet', field: 'category' }],
```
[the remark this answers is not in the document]
rma4ok: Yeah moved it to criteria and got rid of `childrenNodes`

src/MemoryTable.js
rma4ok: There is a lot of code that has data just in a promise and relies on that. This allows us to render the table before the promise resolves. And when it does it just refreshes the tree and updates the table

src/MemoryTable.js
rma4ok: I don't there is a need for the `memoryClient` of this kind, we actually want the hook that produces the tree. Also, this approach could produce a new tree every time, that we are trying to avoid

src/MemoryTable.js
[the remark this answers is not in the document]
daedalus28: Why are we using `disableAutoUpdate`? That means inline filters won't run right away, but there's no reason not to have them update immediately, right?

src/exampleTypes/ResultTable/TableBody.js
```diff
@@ -35,14 +35,15 @@ let TableBody = ({
               {...{ fields, visibleFields, hiddenFields }}
             >
               {_.map(
-                ({ field, display = x => x, Cell = Td }) => (
+                ({ field, display = x => x, Cell = Td, cellProps }) => (
```
daedalus28, line 38: The idea of overriding `Cell` was so we wouldn't need to allow cellProps to be overridden since you can just override Cell and add whatever props you wanted. I'm not completely opposed to adding it, but I wonder if it's needed

src/exampleTypes/ResultTable/index.js
```diff
@@ -123,7 +124,7 @@ let ResultTable = ({
             }}
           />
         </Table>
-        {node.pageSize > 0 && (
+        {!hideFooter && node.pageSize > 0 && (
```
daedalus28, line 127: What use case do we have for hiding the footer? If I recall, it only shows up if there are multiple pages

daedalus28 (changes requested):

src/exampleTypes/ResultTable/TableBody.js
stellarhoof: I'm +1 on `cellProps` since `defaultProps` creates a wrapper. Same argument as autoform really. It's more convenient to have `props` than `component: defaultProps(...)(Input)`

src/exampleTypes/ResultTable/index.js
rma4ok: Yeah it shows the footer for the results count

src/MemoryTable.js
stellarhoof: After talking, we decided that although it's nicer to make this a bit more composable, we don't need the flexibility for now.

src/MemoryTable.js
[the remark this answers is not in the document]
stellarhoof: Does `disableAutoUpdate` prevents inline filters from working?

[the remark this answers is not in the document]
rma4ok: They work, but you need to click `Search` button yourself. Just clicking on options in filters wont apply them

src/exampleTypes/ResultTable/TableBody.js
rma4ok: I've actually already removed `cellProps` in favor of `Cell`. Just to keep the props footprint simpler. Having `Cell: defaultProps(...)(Td)` makes code only couple chars longer in this case

src/MemoryTable.js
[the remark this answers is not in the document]
rma4ok: Good call. All places that define criteria would use <SearchFilters> as well. 
I see now how it makes sense to have it at `denounce: 0` and not to rely on the search buttons to be present.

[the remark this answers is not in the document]
rma4ok: Fixed that

stellarhoof (approved):

src/MemoryTable.js
```diff
@@ -1,42 +1,67 @@
...
-  })
-
-let MemoryTable = ({ data, fields, debug, include, ...props }) => {
-  let service = memoryService(data, { schema: 'data', debug })
-  let [tree] = React.useState(
-    ContextureMobx({ service })({
+export let useMemoryTree = ({
+  records,
+  debug,
+  resultNode = {
```
daedalus28, line 12: this should be `resultsNode`

src/MemoryTable.js
```diff
@@ -1,42 +1,67 @@
...
+    pageSize: 50,
+  },
+  criteriaNodes = [],
```
daedalus28, line 15: let's do a _.castArray so you can just pass one node

daedalus28 (changes requested):

PR #485, opened 2021-03-25
Date range change

Akrion:
- Remove the node update which forces the filter when in popover to close due to the update of the node/tree.
   There also is no need to update the tree node since we are switching between `exact` and `rolling` without an actual date range change.
- Since we do not update the node anymore we use the `useEffect` hook to set the range in the edge case of the `clear/reset` where the node is updated outside of the component.

Note this is version `2.42.4` and we have another PR with the previous version not merged yet.

### DIFF
--- a/src/exampleTypes/Date.js
+++ b/src/exampleTypes/Date.js
@@ -48,7 +48,9 @@ let DateComponent = ({
   excludeRollingRanges = [],
   theme: { DateInput, RadioList, Select },
 }) => {
-  let [range, setRange] = React.useState(node.range !== 'exact' ? 'rolling': 'exact')
+  let [range, setRange] = React.useState(
+    node.range !== 'exact' ? 'rolling' : 'exact'
+  )
 
   // We need the hook to deal with the `clear/reset` case where the node.range is changed but the range remains unchanged
   React.useEffect(() => {
--- a/src/exampleTypes/Date.js
+++ b/src/exampleTypes/Date.js
@@ -48,12 +48,14 @@ let DateComponent = ({
   excludeRollingRanges = [],
   theme: { DateInput, RadioList, Select },
 }) => {
-  let [range, setRange] = React.useState(node.range === 'exact' ? 'exact' : 'rolling')
+  let [range, setRange] = React.useState(
+    node.range === 'exact' ? 'exact' : 'rolling'
+  )
 
   // We need the hook to deal with the `clear/reset` case where the node.range is changed but the range remains unchanged
   React.useEffect(() => {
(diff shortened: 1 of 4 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


Take a look at this commit to see what happened in detail: 4fec046589acec9ca58003e0c1883fe3c53f7680


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 3e800103a022627751b91b0221a9dd7548187648
</p>

src/exampleTypes/Date.js
```diff
@@ -57,30 +66,21 @@ let DateComponent = ({
...
-        value={node.range !== 'exact' ? 'rolling' : 'exact'}
+        value={range}
         style={{ marginBottom: 10 }}
-        onChange={value => {
-          tree.mutate(
-            node.path,
-            value === 'exact'
-              ? { range: 'exact', from: null, to: null }
-              : { range: '', from: null, to: null }
-          )
```
daedalus28, line 68: Don't you still need to mutate? otherwise, you're not modifying the node itself

daedalus28 (changes requested):

src/exampleTypes/Date.js
Akrion: You only need to mutate when you select the actual **date range** not when you switch from `exact` to `rolling`. 
When you select `exact` and chance any of the `to` and `from` we **mutate** ... same with when you select `rolling` and then select the range from the `select` box .... so no you do not have to in the act of selecting the type of range. That was the main issue here.

daedalus28: But if you go from a specific date range in exact to rolling, you'll still have a filtered date range even though it's no longer visible. Likewise if you have a rolling range and then switch to non rolling, it will appear like there's no filter value even though you're still filtering by some range.

Akrion: The UI issue where the values persisted between the `exact/rolling` switch should be now resolved. We do not need to update the underlining node when we switch the type. Only when we select an actual value for the `rolling` range or `to/from` for the exact one.

PR #496, opened 2021-04-23
Introduce addTags

Akrion:
The issue identified by @daedalus28 recently is that we call the `tree.mutate` on every single tag when someone does copy/paste of 100K tags. We limit to only 100 now ... but we still call `tree.mutate` 100 times. 

- Fix the `splitCommas` not being properly passed down from `TagsQuery` to `TagsInput`
- Refactor the logic so that we now have `addTags` method which adds array
- Refactor the common logic into `utils` so that the ExpandableTagsInput also uses it

### DIFF
--- a/src/exampleTypes/TagsQuery/index.js
+++ b/src/exampleTypes/TagsQuery/index.js
@@ -63,7 +63,7 @@ let TagsQuery = ({
             onAddTag(tagObject)
             return tagObject
           }}
-          // onTagsAdded: Called when the all the tags are sanitized and converted to objects with their distance etc.
+          // onTagsAdded: Called when all the tags are sanitized and converted to objects with their distance etc.
           onTagsAdded={tags => tree.mutate(node.path, { tags: [ ...node.tags,...tags ] })}
           removeTag={tag => {
             tree.mutate(node.path, {
--- a/src/exampleTypes/TagsQuery/index.js
+++ b/src/exampleTypes/TagsQuery/index.js
@@ -58,16 +58,11 @@ let TagsQuery = ({
           sanitizeTags={sanitizeTags}
           wordsMatchPattern={wordsMatchPattern}
           tags={_.map(tagValueField, node.tags)}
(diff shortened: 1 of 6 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 392b1e3db21fc0dfc4ad8589300fa3992ff6f888
</p>

src/exampleTypes/TagsQuery/index.js
```diff
@@ -53,16 +54,20 @@ let TagsQuery = ({
...
           tags={_.map(tagValueField, node.tags)}
+          // addTag: Called on every tag being added via the input. Use to decorate the tag into object form
           addTag={tag => {
-            tree.mutate(node.path, {
-              tags: [...node.tags, { [tagValueField]: tag, distance: 3 }],
-            })
-            onAddTag(tag)
+            let tagObject = { [tagValueField]: tag, distance: 3 }
+            onAddTag(tagObject)
+            return tagObject
```
daedalus28, line 65: I think we can simplify this and just have a single addTags method, since both methods end up calling onTagsAdd anyway

daedalus28 (changes requested):
I think we also need to modify the expandable tags input

src/exampleTypes/TagsQuery/index.js
Akrion: Refactored to use one method `addTags`.

Akrion:
> I think we also need to modify the expandable tags input

I also updated that as well as refactored the common fns.

daedalus28 (approved):

PR #498, opened 2021-04-26
Fix the tagsText and more changes to TagsInput

Akrion:
- TagsText is impacted by the `addTags` change
- Default `addTags` to `_.identity`


stellarhoof (approved):
Tested on our product side. The only place where we use this prop directly is on autoform so that needs to be fixed. All the other places use `TagsInput` indirectly and don't directly use this prop.

Also @daedalus28 The previous fix was technically a breaking change.

ccalbaugh (approved):

dshishkov (approved):

PR #501, opened 2021-04-30
Fix maxTags checking

Akrion:
- Better handle the `maxTags` scenario

### DIFF
--- a/src/exampleTypes/TagsQuery/index.js
+++ b/src/exampleTypes/TagsQuery/index.js
@@ -73,7 +73,7 @@ let TagsQuery = ({
               allTags = _.take(maxTags, allTags)
               onTagsDropped(maxTags, tags)
             }
-            tree.mutate(node.path, { tags:  allTags })
+            tree.mutate(node.path, { tags: allTags })
             onAddTag(tags)
           }}
           removeTag={tag => {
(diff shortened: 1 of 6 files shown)


stellarhoof (commented):
I don't follow. Why is the max tags handling being moved from TagsInput to TagsQuery?

Akrion:
> I don't follow. Why is the max tags handling being moved from TagsInput to TagsQuery?

The way we had it before was not correct since no matter what you do in terms of tags sanitizing etc we simply appended what was before with what was just added ... so `max` was broken if you add the tags in multiple copy/paste etc. 

With the current approach, we always limit the total tags and those are then stamped on the tree. Also, `TagsQuery` is the one now calling the `onTagsDropped` and there was no need to pass that and `maxTags` down to `TagsInput` ... We also only care about the `maxTags` limit in the context of the `TagsQuery` really ... in the other cases (unless I am missing something)  we do not really need it.

src/exampleTypes/TagsQuery/index.js
```diff
@@ -67,7 +67,13 @@ let TagsQuery = ({
               tag => ({ [tagValueField]: tag, distance: 3 }),
               tags
             )
-            tree.mutate(node.path, { tags: [...node.tags, ...tagObjects] })
+            let allTags = [...node.tags, ...tagObjects]
```
daedalus28, line 70: Nit: if this was named `tags`, you could use object shorthand in the mutate call below

Akrion: done

PR #504, opened 2021-06-24
[feature] Inline filter buttons

rma4ok:
* [x] `FilterButtonList`: adopt `FilterList` functionality
* [x] Stories: added `FilterButtonList`

### DIFF
--- a/src/FilterButtonList.js
+++ b/src/FilterButtonList.js
@@ -40,7 +40,7 @@ let FilterButtonItem = _.flow(
             F.on(modal)()
             tree.mutate(node.path, { paused: false })
           }}
-          theme={{Button: FilterButton}}
+          theme={{ Button: FilterButton }}
         >
           {title}
         </CheckButton>
@@ -61,7 +61,7 @@ let FilterButtonItem = _.flow(
                 }}
               />
             </div>
-            <Flex style={{justifyContent: "flex-end"}}>
+            <Flex style={{ justifyContent: 'flex-end' }}>
               <Button onClick={() => tree.clear(node.path)}>
                 Clear Filter
               </Button>
@@ -71,7 +71,7 @@ let FilterButtonItem = _.flow(
                   F.off(modal)()
                   tree.mutate(node.path, { paused: true })
                 }}
-                style={{marginLeft: '10px'}}
+                style={{ marginLeft: '10px' }}
               >
                 Done
               </Button>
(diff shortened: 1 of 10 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


Take a look at this commit to see what happened in detail: b9233cf52fb8559d9f00bd0b46a6ab91fae5026b


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against ced354b0940e1c7988e2f8bb029fe4084d25d8ed
</p>

src/purgatory/CheckButton.js
```diff
@@ -22,7 +22,7 @@ let CheckButton = ({
         onChange={_.noop} // prevent another react warning when `checked` is passed but `onChange` isn't
         disabled
       />
-      {children}
+      &nbsp; {children}
```
ccalbaugh, line 25: 👍

src/FilterButtonList.js
```diff
@@ -1,40 +1,51 @@
...
       _.get('label', mappedProps) ||
+      _.get([node.key, 'label'], fields) ||
       _.get([node.field, 'label'], fields) ||
       node.field ||
       node.key
     let description = _.get('description', mappedProps)
     return (
-      <>
-        <CheckButton checked={node.hasValue} onClick={F.on(modal)}>
+      <div>
```
ccalbaugh, line 35: Why change from the fragment here?

ccalbaugh (approved):
One question, just for curiosity really.  Looks good 👍

src/FilterButtonList.js
rma4ok: We need the buttons to be in containers so they are aligned properly with the parent's flex

src/FilterButtonList.js
```diff
@@ -50,57 +61,99 @@ let FilterButtonItem = _.flow(
...
-          nested
-          {...{
-            tree,
-            node: child,
-            fields,
-            mapNodeToProps,
-            className,
-          }}
-        />
+let FilterButtonList = observer(
```
daedalus28, line 97: Let’s move the observable call to the flow. If we don’t, the displayName won’t be automatically picked up and we’d have to set it manually.

src/FilterList.js
```diff
@@ -110,7 +111,11 @@ export let Label = _.flow(
       }}
     >
       <span {...props}>
-        {children || _.get([field, 'label'], fields) || field || ''}
+        {children ||
+          _.get([key, 'label'], fields) ||
```
daedalus28, line 115: Why would we check key? Keys are arbitrary and not likely to be in the schema 🤔

src/FilterButtonList.js
```diff
@@ -50,57 +61,99 @@ let FilterButtonItem = _.flow(
...
+    fields = {},
+    mapNodeToProps = _.noop,
+    allowDuplicateFields = false,
+    className = '',
+    nested = false,
+    style,
+    children,
+    theme: { Icon, Button, FilterButton = Button },
+  }) => {
+    let options = fieldsToOptions(fields)
```
daedalus28, line 110: This and the lines below are likely duplicated from the regular filter list - we should probably factor out a reusable utility to get field options that takes allow duplicates as a flag)

src/FilterButtonList.js
```diff
@@ -50,57 +61,99 @@ let FilterButtonItem = _.flow(
...
+                fields,
+                mapNodeToProps,
+                className,
+              }}
+            />
+          )
+        }, _.get('children', node))}
+
+        {!nested && (
+          <div>
```
daedalus28, line 142: I’m pretty sure we already have cases where we have a filter adder (either in storybooks or in spark). Did you consider those existing cases? We’ll need to update them if filter button list packages up a filter adder

daedalus28 (changes requested):

src/FilterButtonList.js
rma4ok: Yes, good idea. I will move `contexturify` to the definition because this component is using itself for rendering

src/FilterList.js
rma4ok: Our fields are not just the schema, it's schema + overrides where we specify labels by the key. Think of several facets for the same field (title contains/doesn't)

daedalus28: We typically handle this with mapNodeToLabel. The problem with coding against keys is that they aren't guaranteed to be present since you can generally remove and re-add nodes. In your example, someone can remove title contains and does not contain and then add them back. FilterList generally blocks duplicate fields, but query builder does not. 

Also, having schemas have fields that are keys but not fields seems like a recipe for confusion.

src/FilterButtonList.js
rma4ok: Yeah, I've added an optional `addFilters` prop

src/FilterList.js
rma4ok: I see that makes sense now. I will reverse that and add mapNodeToLabel for filter buttons

daedalus28 (approved):

PR #507, opened 2021-07-01
Facet UI changes to limit number of checked items

Akrion:
- Sets a soft limit of 250 checked items at which point we display a note to the user:

CleanShot 2021-07-01 at 22 35 52@2x

- `Select All` is hidden if the visible items are more than 500. `Select All` is visible if all the items are checked to allow the user to uncheck all

This currently does not change the old searches values to only send 500. My thought process here was to limit that at the CES if we want to.

### DIFF
--- a/src/utils/facet.js
+++ b/src/utils/facet.js
@@ -22,7 +22,7 @@ export let Cardinality = _.flow(
   setDisplayName('Cardinality'),
   observer
 )(({ node, tree }) => {
-  let size = _.getOr(10, 'size', node)
+  let size = node.size || 10
   let count = _.get('context.cardinality', node)
   if (count) {
     return (
--- a/src/utils/facet.js
+++ b/src/utils/facet.js
@@ -21,61 +21,77 @@ export let displayBlankFn = () => <i>Not Specified</i>
 export let Cardinality = _.flow(
   setDisplayName('Cardinality'),
   observer
-)(({ node, tree }) =>
-  _.get('context.cardinality', node) ? (
-    <Flex
-      className="contexture-facet-cardinality"
-      justifyContent="space-between"
-    >
-      <div>
-        Showing{' '}
-        {toNumber(_.min([node.size || 10, _.size(node.context.options)]))} of{' '}
-        {toNumber(node.context.cardinality)}
-      </div>
-      {node.context.cardinality > (node.size || 10) && (
+)(({ node, tree }) => {
+  let size = _.getOr(10, 'size', node)
+  let count = _.get('context.cardinality', node)
+  if (count) {
+    return (
+      <Flex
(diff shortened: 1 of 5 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  
<tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


Take a look at this commit to see what happened in detail: e2c9dc63ef569a5767bd3069b4dc587feebed6eb


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 2ac5ca1116d432a3f8038686f060eb19fd28addc
</p>

src/utils/facet.js
```diff
@@ -21,61 +21,67 @@ export let displayBlankFn = () => <i>Not Specified</i>
...
-      justifyContent="space-between"
-    >
-      <div>
-        Showing{' '}
-        {toNumber(_.min([node.size || 10, _.size(node.context.options)]))} of{' '}
-        {toNumber(node.context.cardinality)}
-      </div>
-      {node.context.cardinality > (node.size || 10) && (
+)(({ node, tree }) => {
+  let size = _.getOr(10, 'size', node)
```
daedalus28, line 25: This is a slightly different behavior - before, size would be 10 if it was set to 0, but this will change the behavior when you pass in 0 on purpose. Was this change intentional?

daedalus28 (changes requested):

src/utils/facet.js
Akrion: PAssing size 0 on purpose did not make sense to me... I changed it back.

src/utils/facet.js
```diff
@@ -21,61 +21,67 @@ export let displayBlankFn = () => <i>Not Specified</i>
...
   )
-  let allSelected = _.isEmpty(missingOptions)
-  return (
+  let isAllChecked = _.isEmpty(notChecked)
+  let allChecked = [...node.values, ...notChecked]
+  let isOverTheLimit = _.size(allChecked) > maxChecked
+
+  // If the items are all already selected and we are not going to be over the max if we select all
+  // then show the "Select All". This way we still allow the user to be able to "Unselect all"
+  return !isOverTheLimit || isAllChecked ? (
```
stellarhoof, line 68: I think we can hide it even if they're all checked, since clearing the filter is equivalent to unselecting all.

stellarhoof (approved):

PR #510, opened 2021-07-15
[fix] TagsInput: rendering new tag input before the list of tags

rma4ok:
* [x] TagsInput: rendering new tag input before the list of tags


stellarhoof:
What does this solve?

PR #524, opened 2021-09-22
Fix advanced search node wrap

stellarhoof:
Fix wrapping a node in an AND/OR group when in advanced search.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "contexture-react",
-  "version": "2.49.4",
+  "version": "2.49.5",
   "description": "React components for building contexture interfaces",
   "main": "dist/index.js",
   "scripts": {
@@ -94,6 +94,7 @@
     "babel-jest": "^23.4.2",
     "babel-loader": "^8.0.6",
     "babel-plugin-require-context-hook": "^1.0.0",
+    "chokidar": "^3.5.2",
     "contexture": "^0.5.2",
     "contexture-client": "^2.31.2",
     "contexture-elasticsearch": "^0.18.2",
--- a/src/utils/search.js
+++ b/src/utils/search.js
@@ -30,22 +30,21 @@ export let transformNodeFromField = args => node => ({
   ...newNodeFromField(args),
 })
 
-export let indent = (Tree, parent, node, skipDefaultNode) => {
+export let indent = async (Tree, parent, node, skipDefaultNode) => {
   // Reactors:
   //   OR -> And, nothing
   //   AND -> OR, others if has value
   //   to/from NOT, others if has value
   let join = (parent || node).join
-  let key = node.key
   let path = _.toArray(node.path)
-
-  Tree.mutate(path, { key: `${key}-${join}-group` })
-  Tree.wrapInGroup(path, {
-    key,
+  let wrapperNode = {
+    key: `${node.key}-${join}-group`,
     join: oppositeJoin(join),
-  })
-  if (!skipDefaultNode) Tree.add(path, blankNode())
-  return Tree.getNode(path)
+  }
+  let wrapperPath = [..._.dropRight(1, path), wrapperNode.key]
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -94,6 +94,7 @@
     "babel-jest": "^23.4.2",
     "babel-loader": "^8.0.6",
     "babel-plugin-require-context-hook": "^1.0.0",
+    "chokidar": "^3.5.2",
```
stellarhoof, line 97: This is needed for `npm run watch` to work.

src/utils/search.js
```diff
@@ -30,22 +30,21 @@ export let transformNodeFromField = args => node => ({
...
+export let indent = async (Tree, parent, node, skipDefaultNode) => {
   // Reactors:
   //   OR -> And, nothing
   //   AND -> OR, others if has value
   //   to/from NOT, others if has value
   let join = (parent || node).join
-  let key = node.key
   let path = _.toArray(node.path)
-
-  Tree.mutate(path, { key: `${key}-${join}-group` })
```
daedalus28, line 42: this is a good catch - I don't think we've ever had a case to `mutate` a `key`, and I'm honestly not sure if it worked 🤔

src/utils/search.js
```diff
@@ -30,22 +30,21 @@ export let transformNodeFromField = args => node => ({
   ...newNodeFromField(args),
 })
 
-export let indent = (Tree, parent, node, skipDefaultNode) => {
+export let indent = async (Tree, parent, node, skipDefaultNode) => {
```
daedalus28, line 33: I'm not sure this needs to be async - see below.

Also as @re1ro and I discussed, this seems like something we should consider moving into `contexture-client`. The only reason it's here was because of the `blankNode` stuff, but that seems like the canonical thing to do when indenting anyway. The client has generally been ignorant of schemas, but that causes stuff like this to be orphaned and live in contexture-react instead.

daedalus28 (changes requested):

src/utils/search.js
```diff
@@ -30,22 +30,21 @@ export let transformNodeFromField = args => node => ({
...
-    key,
+  let wrapperNode = {
+    key: `${node.key}-${join}-group`,
     join: oppositeJoin(join),
-  })
-  if (!skipDefaultNode) Tree.add(path, blankNode())
-  return Tree.getNode(path)
+  }
+  let wrapperPath = [..._.dropRight(1, path), wrapperNode.key]
+  await Tree.wrapInGroup(path, wrapperNode)
```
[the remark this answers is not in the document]
stellarhoof: There are no true async mutations in the client as far as I can see, however, a bunch of fns are marked as `async`.  [wrapInGroup](https://github.com/smartprocure/contexture-client/blob/b3d831cb004d06dfdd7a3b3d8d29fa8503834df0/src/actions/wrap.js#L38), [replace](https://github.com/smartprocure/contexture-client/blob/b3d831cb004d06dfdd7a3b3d8d29fa8503834df0/src/actions/index.js#L90) isn't but [add](https://github.com/smartprocure/contexture-client/blob/b3d831cb004d06dfdd7a3b3d8d29fa8503834df0/src/actions/index.js#L28) is as well. I'm awaiting here on the off chance that the calls to `add` and `getNode` fail due to timing issues. I could also make a PR to remove the `async` wrapping in contexture-client instead. I just assumed they were added for a reason.

src/utils/search.js
stellarhoof: See my comment above for the async stuff.

> The client has generally been ignorant of schemas, but that causes stuff like this to be orphaned and live in contexture-react instead.

This function is not aware of schemas though.

src/utils/search.js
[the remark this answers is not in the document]
stellarhoof: Removed `async`

PR #537, opened 2022-01-05
Feature/search data testid

nico-olivares:
-Added two data-testid. One to the main search bar and one to the main search button.
-Added three data-path. Facet, date, and number. (per discussion with Alejandro)
-Updated the package.json (version number)
-Updated the package-lock.json (version number)
-Updated the changelog.mdx

### DIFF
--- a/src/exampleTypes/Date.js
+++ b/src/exampleTypes/Date.js
@@ -56,7 +56,7 @@ let DateComponent = ({
   )
 
   return (
-    <div>
+    <div data-path={node.path}>
       <RadioList
         options={F.autoLabelOptions(['exact', 'rolling'])}
         value={node.range !== 'exact' ? 'rolling' : 'exact'}
--- a/src/exampleTypes/ResultTable/index.js
+++ b/src/exampleTypes/ResultTable/index.js
@@ -89,8 +89,8 @@ let ResultTable = ({
   }
 
   return (
-    <div data-path={node.path}>
-      <Table>
+    <>
+      <Table data-path={node.path}>
         <Thead>
           <Tr>
             {F.mapIndexed(
@@ -138,7 +138,7 @@ let ResultTable = ({
           }}
         />
       )}
-    </div>
+    </>
   )
 }
 
--- a/src/exampleTypes/ResultTable/index.js
+++ b/src/exampleTypes/ResultTable/index.js
@@ -89,7 +89,7 @@ let ResultTable = ({
   }
 
   return (
-    <>
+    <div data-path={node.path}>
       <Table>
         <Thead>
           <Tr>
@@ -138,7 +138,7 @@ let ResultTable = ({
           }}
         />
       )}
-    </>
+    </div>
   )
 }
 
--- a/src/exampleTypes/TagsQuerySearchBar.js
+++ b/src/exampleTypes/TagsQuerySearchBar.js
@@ -79,7 +79,11 @@ let SearchBar = ({
           e.stopPropagation()
         }}
       >
-        <Box style={searchBarBoxStyle} onClick={F.off(collapse)}>
+        <Box
+          data-testid="main-search-bar"
+          style={searchBarBoxStyle}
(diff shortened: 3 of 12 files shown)


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 08f104935787aab9b22071717bc34039b448714d
</p>

src/exampleTypes/TagsQuerySearchBar.js
```diff
@@ -29,6 +29,7 @@ let buttonStyle = {
 
 let AnimatedButton = ({ disabled, style, className, ...props }) => (
   <Button
+    data-testid="button-main-search"
```
stellarhoof, line 32: Instead why not just add the same `data-path` to `SearchBar` and select the input and the search button from there?

stellarhoof (commented):
If we're adding `data-path` to some example types we should add it to all the example types components to be consistent. Otherwise we'll have to make follow-up PRs to allow selecting other types of filters.

warrenv (approved):

nico-olivares:
> If we're adding `data-path` to some example types we should add it to all the example types components to be consistent. Otherwise we'll have to make follow-up PRs to allow selecting other types of filters.

@stellarhoof I agree, but when I did something like that before I got push back for doing too many at once.
Don't you think it would make sense to add a `data-path` to any component that receives `node`?

stellarhoof:
> Don't you think it would make sense to add a `data-path` to any component that receives `node`

Not necessarily. There are components that have access to `node` that are not "example types". Do not worry about the word "example". "Type" refers to the type of a node (`tagsQuery`, `facet`, `number`, etc...). Each type has a component so I don't see adding `data-path` to each example type as overreaching because we know there is simply no good way of selecting filters.

`TagsQuerySearchBar` is not actually a component for a node type but I think it also makes sense to add a `data-path` because there's no good way of selecting it right now.

nico-olivares:
@stellarhoof take a second look. I got rid of the data-testids and added a single data-path to the whole component like you suggested. I also added a couple of other data-paths to what I saw made sense. Let me know if there's any I should get rid of or add.

src/exampleTypes/ResultTable/index.js
```diff
@@ -89,7 +89,7 @@ let ResultTable = ({
   }
 
   return (
-    <>
+    <div data-path={node.path}>
```
stellarhoof, line 92: Tables are a bit delicate. I don't know if wrapping this one on a div will wreak havoc somewhere. You can pass `data-path` to the `Table` component instead.

nico-olivares: @stellarhoof I moved it to the `Table`

stellarhoof (approved):

nico-olivares:
@daedalus28 could you please review / merge? I don't know who else has merging permits here. Thanks,

src/exampleTypes/TagsQuerySearchBar.js
nico-olivares: Done.

dubiousdavid (approved):
Looks good.

src/exampleTypes/Date.js
```diff
@@ -56,7 +56,7 @@ let DateComponent = ({
   )
 
   return (
-    <div>
+    <div data-path={node.path}>
```
daedalus28, line 59: These changes might be better generically as part of the contexturify HOCs we have so we don't have to remember to do this on every component

PR #582, opened 2023-01-16
Remove HOCs

stellarhoof:
- Remove usage of `withNode`, `withTheme`, `withLoader`, `withTreeLens`, `contexturify`, and `contexturifyWithoutLoader` and mark them as deprecated
- Introduce two new hooks `useNode` and `useTheme`, which are very simple and cover all the functionality provided by the deprecated ones
- Add `Loader` component to the theme
- Remove usage of mobx's `inject`
- Drag N Drop HOCs will be replaced later
- Remove `react-recompose`


decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:no_entry_sign:</td>
    </tr>
  
<tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>


<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>:exclamation: This PR is BIG (+2106 -1937)  
Please keep it below 500 net changes</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 1616ac029754a6bf54edac8a7d52473fd0e07bfa
</p>

decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>:exclamation: This PR is BIG (+1762 -1638)  
Please keep it below 500 net changes</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 90adb94ece2423027fce542ded2f69f3c7597c8d
</p>

decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>:exclamation: This PR is BIG (+1763 -1639)  
Please keep it below 500 net changes</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against aa35aa20ee164c9f099f482af62c8457671ba608
</p>

decrapifier:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>:exclamation: This PR is BIG (+1765 -1641)  
Please keep it below 500 net changes</td>
    </tr>
  </tbody>
</table>



<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/js">dangerJS</a> against 63e1071235dc8fb58d13f56278ab3a009e47d282
</p>

src/utils/hooks.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import ThemeContext from './theme/context.js'
+
+export const useNode = (node, path, tree) => {
+  node = node || (tree && path && tree.getNode(path))
+  if (!node) throw Error(`Node not provided, and couldn't find node at ${path}`)
+  return node
+}
```
daedalus28, line 8: `useNode` definitely needs to be an implementation like the one I've got WIP using the client watchNode API. If you're dropping observers and aren't triggering rerenders from the hook, this PR means we're relying on the top level observer and forcing the entire UI to rerender.

Also, this migration needs to be piecemeal since its such a huge change

stellarhoof: Not sure I follow. `withNode` never triggered any re-renders. Here's `contexturify`

```javascript
export let contexturify = _.flow(observer, withLoader, withNode, withTheme)
```

`observer` is applied first so any property dereferenced in the other HOCs are not tracked.

> `useNode` definitely needs to be an implementation like the one I've got WIP using the client watchNode API.

I think this comment was made with the assumption that the current `withNode` hoc re-renders components, but it does not, so this `useNode` hook is not equivalent to your watchNode API.

stellarhoof: > Also, this migration needs to be piecemeal since its such a huge change

It touches lots of files but they're all the same change. Perhaps we can get together to go over it.

daedalus28: Ah I missed the part in the diff where you're adding observer to every component and just focused on dropping contexturify. As long as each component is still an observer, there's no issue 👍 

With that said, it does seem like we might not want to make this change given that we're going to move off observer as well with a useNode hook that also does subscription. This PR gets us closer to that end goal though, so I'm not opposed 🤔

daedalus28:
One more thing on this PR - I'm not sure removing the HOC and inlining loaders everywhere is net gain, especially since we also want to add generic hover to trigger a `focus`/`hover` client state. We probably still want a way to provide off the shelf contexture wrapping for those things - a HOC actually seems like a good solution there. Open to alternatives, but duplicating the logic everywhere seems worse to me.

stellarhoof:
With respect to loaders, I actually prefer the duplication because:

1. We can customize where to put the loader. Perhaps we wanna be a bit fancier and only wrap an input instead of the entire component
2. Less re-renders: a hoc will re-render the wrapped component if it re-renders itself, whereas a wrapper won't. I have not tested this, but I think it'll get rid of our problem with double rendering when loaders switch state.
3. In general, hocs are harder to reason about (at least for me)

I'm not sure how the focus/hover feature would work so I can't comment there.

